PR #3, opened 2017-05-24
Add MSMQ

bording:
This takes everything that was removed from Core in https://github.com/Particular/NServiceBus/pull/4746 and adds it here, except for the legacy distributor stuff, which was removed altogether.

This won't build yet because it's referencing local packages I built. After https://github.com/Particular/NServiceBus/pull/4746  is merged and an alpha build is up on MyGet, I'll update this to use them instead.


src/NserviceBus.Transport.Msmq/InstanceMapping/InstanceMappingFileFeature.cs
```diff
@@ -25,7 +27,7 @@ protected override void Setup(FeatureConfigurationContext context)
             }
 
             var checkInterval = context.Settings.Get<TimeSpan>(CheckIntervalSettingsKey);
-            var endpointInstances = context.Routing.EndpointInstances;
+            var endpointInstances = context.Settings.Get<EndpointInstances>();
```
bording: @Particular/nservicebus-maintainers Do we want to consider making `Routing` public? I was able to work around not having it.

src/NserviceBus.Transport.Msmq/MsmqTimeToBeReceivedOverrideCheck.cs
```diff
@@ -19,8 +19,19 @@ public StartupCheckResult CheckTimeToBeReceivedOverrides()
             var isTransactional = settings.GetRequiredTransactionModeForReceives() != TransportTransactionMode.None;
             var outBoxRunning = settings.IsFeatureActive(typeof(Features.Outbox));
 
-            var messageAuditingConfig = settings.GetOrDefault<AuditConfigReader.Result>();
-            var auditTTBROverridden = messageAuditingConfig?.TimeToBeReceived > TimeSpan.Zero;
+
+            //var messageAuditingConfig = settings.GetOrDefault<AuditConfigReader.Result>();
+            //var auditTTBROverridden = messageAuditingConfig?.TimeToBeReceived > TimeSpan.Zero;
+
+            //workaround for AuditConfigReader.Result not being accessible
```
bording: @Particular/nservicebus-maintainers `AuditConfigReader.Result` is internal, so this was a bit of a hack to get the info needed here. Maybe we should expose a way to get to the configured audit TTBR value?

src/NserviceBus.Transport.Msmq/MsmqTransport.cs
```diff
@@ -37,7 +38,7 @@ public override TransportInfrastructure Initialize(SettingsHolder settings, stri
                 throw new Exception("Faults forwarding requires an error queue to be specified using 'EndpointConfiguration.SendFailedMessagesTo()'");
             }
 
-            settings.EnableFeature(typeof(InstanceMappingFileFeature));
+            settings.Set(typeof(InstanceMappingFileFeature).FullName, FeatureState.Enabled);
```
bording: @Particular/nservicebus-maintainers Any concern about this workaround? `EnableFeature` is internal.

src/NServiceBus.Transport.Msmq.Tests/InstanceMapping/InstanceMappingFileMonitorTests.cs
```diff
@@ -70,7 +70,7 @@ public void Reload_should_throw_when_file_does_not_exist()
             });
 
             var monitor = new InstanceMappingFileMonitor("unused", TimeSpan.Zero, timer, fileAccess, new EndpointInstances());
-            await monitor.PerformStartup(null);
+            await monitor.Start(null);
```
bording: @Particular/nservicebus-maintainers `PerformStartup` is internal, so I had to add a method to the feature to be able to get the test to work. This seems like on OK workaround, and I'm not thinking we need to do anything here.

src/NServiceBus.Transport.Msmq.Tests/InstanceMapping/InstanceMappingFileFeatureTests.cs
```diff
@@ -11,13 +11,15 @@ public class InstanceMappingFileFeatureTests
         [Test]
         public void Should_configure_default_values()
         {
-            var feature = new InstanceMappingFileFeature();
-            var settings = new SettingsHolder();
+            // Test has to be commented out for now, no way to access the defaults
```
bording: @Particular/nservicebus-maintainers This is the one place where I couldn't figure out a workaround, so I had to just comment out the test for now.

`ConfigureDefaults` is internal, and derived types have no access to `registeredDefaults`, so I couldn't pull off a similar change like I did for `PerformStartup`.

I'm not sure we want to expose this to features, so maybe this test just needs to be deleted?

src/NserviceBus.Transport.Msmq/MsmqTimeToBeReceivedOverrideCheck.cs
indualagarsamy: @andreasohlund - since there's limitations around TTBR for audit messages, do we even care? Should we get rid of the TTBR option for audit messages? 
Also as discussed with @bording, I think the AuditConfigReader will be so much easier after getting rid of the registry setting as well.

bording: I'm not sure we want to 🔥 the feature for everyone just because MSMQ can't support it.

At the very least, if we stored the two audit values in settings as separate values with string keys instead of wrapped in the internal `Result` class, the TTBR value would be readable without `dynamic` trickery.

indualagarsamy: @bording - :) True. That was me and my MSMQ hat! :)

andreasohlund:
@bording https://github.com/Particular/NServiceBus/pull/4746 is now merged

src/NserviceBus.Transport.Msmq/MsmqTimeToBeReceivedOverrideCheck.cs
bording: So my fancy `dynamic` workaround didn't actual work, because `dynamic` still respects visibility. So even though I can see the property on the object, it wouldn't let me access it. 😢

bording:
OK, I've gotten it down to 3 failing tests. One is because of the audit TTBR value not being accessible, so we're definitely going to need to change core to make that available.

The other two are timing out, but I haven't had a chance to find out why yet.

src/NserviceBus.Transport.Msmq/MsmqTimeToBeReceivedOverrideCheck.cs
andreasohlund: @bording I'll make this public in the core and update the PR

src/NserviceBus.Transport.Msmq/MsmqTransport.cs
andreasohlund: Given that this is now in the MSMQ project we can just enable the feature by default  and remove this code?

src/NServiceBus.Transport.Msmq.Tests/InstanceMapping/InstanceMappingFileFeatureTests.cs
```diff
@@ -0,0 +1,25 @@
...
+    using NUnit.Framework;
+    //using Settings;
+
+    [TestFixture]
+    public class InstanceMappingFileFeatureTests
+    {
+        [Test]
+        public void Should_configure_default_values()
+        {
+            // Test has to be commented out for now, no way to access the defaults
```
andreasohlund: Why is this, internalsvisibleto should make them available?

src/NserviceBus.Transport.Msmq/InstanceMapping/InstanceMappingFileFeature.cs
andreasohlund: Feels good enough for now?

src/NserviceBus.Transport.Msmq/MsmqTimeToBeReceivedOverrideCheck.cs
andreasohlund: PR raised to expose this https://github.com/Particular/NServiceBus/pull/4764

src/NServiceBus.Transport.Msmq.Tests/InstanceMapping/InstanceMappingFileFeatureTests.cs
andreasohlund: I see what you mean now. It seems reasonable to me to expose the registered defaults in the core?

I can handle this as a separate PR if needed? cc @Particular/nservicebus-maintainers

andreasohlund:
We also need to add a section about this to https://docs.particular.net/nservicebus/upgrades/6to7/

src/NServiceBus.Transport.Msmq.Tests/InstanceMapping/InstanceMappingFileFeatureTests.cs
bording: Should be possible to unblock this by making `registeredDefaults` protected, assuming we're OK with giving features access to the collection.

src/NserviceBus.Transport.Msmq/MsmqTransport.cs
bording: Oh yeah, good point! I'll make this change.

bording:
OK, we still have the one `InstanceMappingFileFeature` test that is commented out for now until we decide if/how to expose the relevant parts externally, but that can be resolved after this is merged.

So, I believe this is ready to be merged!

PR #25, opened 2017-05-31
Sendonly endpoint and error queue

SeanFeldman:
fixes https://github.com/Particular/NServiceBus/issues/4769

@Particular/msmq-transport-maintainers please review

### DIFF
--- a/src/NServiceBus.Transport.Msmq/MsmqTransport.cs
+++ b/src/NServiceBus.Transport.Msmq/MsmqTransport.cs
@@ -32,7 +32,7 @@ namespace NServiceBus
             Guard.AgainstNull(nameof(settings), settings);
             string errorQueue;
 
-            if (!settings.TryGetExplicitlyConfiguredErrorQueueAddress(out errorQueue) && !settings.Get<bool>("Endpoint.SendOnly"))
+            if (!settings.Get<bool>("Endpoint.SendOnly") && !settings.TryGetExplicitlyConfiguredErrorQueueAddress(out errorQueue))
             {
                 throw new Exception("Faults forwarding requires an error queue to be specified using 'EndpointConfiguration.SendFailedMessagesTo()'");
             }
--- a/src/NServiceBus.Transport.Msmq/MsmqTransport.cs
+++ b/src/NServiceBus.Transport.Msmq/MsmqTransport.cs
@@ -32,7 +32,7 @@ namespace NServiceBus
             Guard.AgainstNull(nameof(settings), settings);
             string errorQueue;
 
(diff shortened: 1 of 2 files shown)


src/NServiceBus.Transport.Msmq/MsmqTransport.cs
```diff
@@ -32,7 +32,7 @@ public override TransportInfrastructure Initialize(SettingsHolder settings, stri
             Guard.AgainstNull(nameof(settings), settings);
             string errorQueue;
 
-            if (!settings.TryGetExplicitlyConfiguredErrorQueueAddress(out errorQueue))
+            if (!settings.TryGetExplicitlyConfiguredErrorQueueAddress(out errorQueue) && !settings.Get<bool>("Endpoint.SendOnly"))
```
indualagarsamy: Wouldn't it be better to check for the SendOnly configuration condition first, and if and only then check the queue configuration? In other words, I think instead of appending to the condition, you should check for the send only endpoint first.

SeanFeldman: Good idea, to shortcut it if it's a readonly endpoint.

indualagarsamy:
@Particular/msmq-transport-maintainers - Shouldn't this change also be in the V6 core / support branch?

indualagarsamy:
@Particular/nservicebus-maintainers - what's the process of fixing MSMQ bugs or features? Should they be done first in the MSMQ transport repo, PR merged and then back-ported to the NSB repo?

bording:
@indualagarsamy Given that NSB's `develop` branch is now v7, we aren't going to be shipping a new v6 minor unless there is a *really* good reason, so there aren't going to be any more new features on v6. Any new MSMQ features would need to be in this repo. That does mean those new features will have to wait for NSB 7, though.

For bug fixes that impact 6.x and v7, I don't believe the order matters right now since v7 is still WIP, so we wouldn't be doing a new release from this repo that we'd want to coordinate releases on.

Once NSB 7 is released, then I'm still not sure it matters, as long as they are both done so we can push out all the releases and announce them at the same time.

indualagarsamy:
@bording - For features I agree. For bugs, I prefer for the fixes to start in this repo, as it will involve the maintainers of this repo. Once the fix has been vetted and merged, it can transition as a patch to the core 6.0 master. Do you see any issues with my proposal? Once NSB 7 is released, we still are on the hook to support version 6.0, until our support policy expires for 6.0, so any bugs which result in a message loss scenario etc, would still need to be back ported.

PR #31, opened 2017-07-31
Using Task.Yield before MessageQueue.Send results in a performance improvement

ramonsmits:
Using `Task.Yield()` before `MessageQueue.Send` results in a significant performance improvement when doing lots of sends.

```c#
IMessageSession messageSession;
var numberOfMessages = 150000;
var tasks = new List<Task>(numberOfMessages);
for (var i = 0; i < numberOfMessages; i++)
{
    tasks.Add(messageSession.Send(new MyCommand()));
}
await Task.WhenAll(tasks).ConfigureAwait(false);
```
Transactional queues:

- Without Task.Yield: 3,498.31 msg/s
- With Task.Yield:12,793.95 msg/s

Non-transactional queues:

- Without Task.Yield: 4,288.47 msg/s
- With Task.Yield: 20,373.91 msg/s

This means that with the `Task.Yield` this executes more than 350% faster sends on transactional queues and 475% faster sends on non-transactional queues.

This PR has 2 commits, the first commit adds the `Task.Yield` to `MsmqMessageDispatcher.Dispatch`, meaning the transport operations as executed sequentially. The 2nd commit moves `Task.Yield` to `MsmqMessageDispatcher.ExecuteTransportOperation` and uses `Task.WhenAll` 

Original PR: https://github.com/Particular/NServiceBus/pull/4880

### DIFF
--- a/src/NServiceBus.Transport.Msmq/MsmqMessageDispatcher.cs
+++ b/src/NServiceBus.Transport.Msmq/MsmqMessageDispatcher.cs
@@ -24,7 +24,7 @@ namespace NServiceBus.Transport.Msmq
             this.messageLabelGenerator = messageLabelGenerator;
         }
 
-        public async Task Dispatch(TransportOperations outgoingMessages, TransportTransaction transaction, ContextBag context)
+        public Task Dispatch(TransportOperations outgoingMessages, TransportTransaction transaction, ContextBag context)
         {
             Guard.AgainstNull(nameof(outgoingMessages), outgoingMessages);
 
@@ -33,15 +33,17 @@ namespace NServiceBus.Transport.Msmq
                 throw new Exception("The MSMQ transport only supports unicast transport operations.");
             }
 
-            await Task.Yield();
+            var transportOperationTasks = new List<Task>(outgoingMessages.UnicastTransportOperations.Count);
 
             foreach (var unicastTransportOperation in outgoingMessages.UnicastTransportOperations)
             {
-                ExecuteTransportOperation(transaction, unicastTransportOperation);
+                transportOperationTasks.Add(ExecuteTransportOperation(transaction, unicastTransportOperation));
             }
+
+            return Task.WhenAll(transportOperationTasks);
         }
 
-        void ExecuteTransportOperation(TransportTransaction transaction, UnicastTransportOperation transportOperation)
+        async Task ExecuteTransportOperation(TransportTransaction transaction, UnicastTransportOperation transportOperation)
         {
             var message = transportOperation.Message;
 
@@ -76,6 +78,8 @@ namespace NServiceBus.Transport.Msmq
                         }
 
                         var label = GetLabel(message);
+
+                        await Task.Yield();
 
                         if (transportOperation.RequiredDispatchConsistency == DispatchConsistency.Isolated)
                         {
--- a/src/NServiceBus.Transport.Msmq/MsmqMessageDispatcher.cs
+++ b/src/NServiceBus.Transport.Msmq/MsmqMessageDispatcher.cs
@@ -70,31 +70,30 @@ namespace NServiceBus.Transport.Msmq
                         toSend.UseJournalQueue = settings.UseJournalQueue;
                         toSend.TimeToReachQueue = settings.TimeToReachQueue;
 
-                        string replyToAddress;
-
-                        if (message.Headers.TryGetValue(Headers.ReplyToAddress, out replyToAddress))
+                        if (message.Headers.TryGetValue(Headers.ReplyToAddress, out var replyToAddress))
                         {
                             toSend.ResponseQueue = new MessageQueue(MsmqAddress.Parse(replyToAddress).FullPath);
                         }
 
                         var label = GetLabel(message);
 
-                        await Task.Yield();
-
-                        if (transportOperation.RequiredDispatchConsistency == DispatchConsistency.Isolated)
+                        await Task.Run(() =>
                         {
-                            q.Send(toSend, label, GetIsolatedTransactionType());
-                            return;
-                        }
 
-                        MessageQueueTransaction activeTransaction;
-                        if (TryGetNativeTransaction(transaction, out activeTransaction))
-                        {
-                            q.Send(toSend, label, activeTransaction);
-                            return;
-                        }
+                            if (transportOperation.RequiredDispatchConsistency == DispatchConsistency.Isolated)
+                            {
+                                q.Send(toSend, label, GetIsolatedTransactionType());
+                                return;
+                            }
 
-                        q.Send(toSend, label, GetTransactionTypeForSend());
+                            if (TryGetNativeTransaction(transaction, out var activeTransaction))
+                            {
+                                q.Send(toSend, label, activeTransaction);
+                                return;
+                            }
+
+                            q.Send(toSend, label, GetTransactionTypeForSend());
+                        }).ConfigureAwait(false);
                     }
                 }
             }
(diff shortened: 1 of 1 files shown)


src/NServiceBus.Transport.Msmq/MsmqMessageDispatcher.cs
```diff
@@ -33,15 +33,17 @@ public Task Dispatch(TransportOperations outgoingMessages, TransportTransaction
...
+            var transportOperationTasks = new List<Task>(outgoingMessages.UnicastTransportOperations.Count);
+
             foreach (var unicastTransportOperation in outgoingMessages.UnicastTransportOperations)
             {
-                ExecuteTransportOperation(transaction, unicastTransportOperation);
+                transportOperationTasks.Add(ExecuteTransportOperation(transaction, unicastTransportOperation));
             }
 
-            return TaskEx.CompletedTask;
+            return Task.WhenAll(transportOperationTasks);
```
bording: If you add something like [this](https://github.com/Particular/NServiceBus.RabbitMQ/blob/develop/src/NServiceBus.Transport.RabbitMQ/Sending/MessageDispatcher.cs#L37), the you can avoid allocating another `Task` with the `WhenAll` call when there is only one message to send.

src/NServiceBus.Transport.Msmq/MsmqMessageDispatcher.cs
```diff
@@ -77,6 +79,8 @@ void ExecuteTransportOperation(TransportTransaction transaction, UnicastTranspor
 
                         var label = GetLabel(message);
 
+                        await Task.Yield();
```
bording: You could move this call to the top of the method instead, and then the entire method body would be executed in the continuation.

ramonsmits: @bording I didn't because there is some validation happening before the Yield. If you choose to do this then do this as early as possible? Is there some optimization by making it the first statement of the method?

bording: It just means that you'll be returning to the calling method that much sooner vs. executing everything before the yield on the calling thread.

ramonsmits: I was more considering the best/optimal usage. For example, you can make it the first statements in a method because you know the code will hit a blocking call as you suggest or you put it just before the blocking call. Not sure when you would want to choose one over the other.

Also, I've read that `Task.Yield` is using context synchronization. You cannot do `.ConfigureAwait(false)` on it and an alternative would be to use `await Task.Run(..).ConfigureAwait(false)` but after testing that it didn't make a difference while the code also becomes more complex due to the lambda.

bording: You make a good point about the synchronization context. If you're testing on a console app, it won't make any difference, but if run from something that does have a synchronization context, that would mean you're resuming back on the calling thread...

Try testing it on a WinForms app or something and see how it looks. I think you might need to change to Task,Run after all.

ramonsmits: I will after I get back from holiday.

indualagarsamy:
@ramonsmits - any updates on this PR?

janovesk:
The missing `ConfigureAwait(false)` for `Task.Yield()` is problematic. It will dead-lock on WinForms/WPF.

While there is [work](https://github.com/dotnet/corefx/issues/23898) [being](https://github.com/dotnet/corefx/issues/15490) [done](https://github.com/dotnet/csharplang/issues/645) to fix the underlying issue with not being able to `ConfigureAwait(false)` on `Task.Yield()`, it's not going to happen anytime soon. 

If we do anything it has to be with `Task.Run()` for now.

Want to update the PR @ramonsmits ?

bording:
@janovesk While that is generally true, in this case I don't believe it would be a problem. You can only deadlock if you're capturing the WinForms/WPF `SynchronizationContext`, and in this case, the code should be called from a thread pool thread, so that won't be an issue here.

ramonsmits:
@bording @janovesk In my code snippet I used `IMessageSession` so I would expect that this could be used from a UI thread.

bording:
> @bording @janovesk In my code snippet I used IMessageSession so I would expect that this could be used from a UI thread.

Ah yes, I was thinking this was on the receiving side, not for sending. `Task.Run` is definitely better then.

danielmarbach:
The dispatcher is used from to entry points. One entry point is from within the pipeline and on entry point is from the message session. Explicitly deciding to offload something is a top level concern and should not be done on the micro level. From the perspective of the pipeline the top level is the transport since the transport is in charge of the "threading" and concurrency. From the perspective of the message session the top level is the user code that calls the message session. Therefore it is the responsibility of the user code to decide whether they want to wrap each session call in a `Task.Run`, or package multiple session into fewer Task.Run or even call those sequentially. By issuing a `Task.Run` inside the dispatcher the dispatcher becomes a liar because it looks like an asynchronous method but it actually offloads to the worker pool. 

> The threadpool is an app-global resource, so you should leave the callers of your library to decide how and when they want to use it. Your library methods should not use the threadpool (Task.Run, Parallel.For, ...) internally in secret. Most of the time, they should have async signatures if and only if their internal implementations are truly async. [see](https://channel9.msdn.com/Series/Three-Essential-Tips-for-Async/Async-Library-Methods-Shouldn-t-Lie)

With this change you might optimize for multiple sends but you are completely removing the ability for the caller to control the concurrency aspects and enforcing high pressure on the threadpool. Furthermore these changes can significantly influence the performance from the receive side of things since the dispatching of messages would enforce offloading although the pump already did offload and would have a dedicated thread for the receive operation already. Essentially it doubles the pressure on the pool per concurrent receive operation.

The dispatcher is a general purpose IO bound interface. That's why it is async and has to be assumed truly async. From an API perspective that returns a task it is totally valid to be execute stuff synchronously. In fact that is by design everything before the first await statement. If you don't have anything to await you'd return `Task.CompletedTask` which means the whole method is synchronous. That's OK. Explicitly offloading on a micro level deep down in a call hierarchy is not.

Feel free to ping me anytime if you need more clarifications on this topic. Btw. I'm always happy to review PRs when it comes to async or TPL or even it it doesnt :)

danielmarbach:
I would like to suggest to update the MSMQ doco with a specific example how to send messages as fast as possible with the message session. Similar to something I did in https://docs.particular.net/nservicebus/handlers/async-handlers#concurrency-small-amount-of-concurrent-message-operations. So in essence the snippet with a slight tweak in the description of this PR can be turned into MSMQ specific documentation

```c#
IMessageSession messageSession;
var numberOfMessages = 150000;
var tasks = new List<Task>(numberOfMessages);
for (var i = 0; i < numberOfMessages; i++)
{
    tasks.Add(Task.Run(messageSession.Send(new MyCommand())));
}
await Task.WhenAll(tasks).ConfigureAwait(false);
```

janovesk:
Yeah, @danielmarbach is right. I agree.

This decision needs to be left up to the end user. I don't think we should implement this as general dispatcher behaviour. 

I'll get a PR with documentation on why the options mentioned in https://docs.particular.net/nservicebus/handlers/async-handlers#concurrency-small-amount-of-concurrent-message-operations makes sense for high-send volume scenarios for MSMQ.

ramonsmits:
@danielmarbach Good explanation although it is contradicting:

> Your library methods should not use the threadpool 

vs

> should have async signatures if and only if their internal implementations are truly async

Which in our case with MSMQ it is not.  It is weird because our implementation isn't truly async that the end-user needs to add this `Task.Run` workaround. That is not what the end user would expect and is only needed with MSMQ.

By the way, is the threadpool also involved when doing `Task.Yield`?

danielmarbach:
@ramonsmits 

Well from the perspective of MSMQ it is contradicting I agree. It is not when you look at the dispatcher component from a design perspective

> The dispatcher is a general purpose IO bound interface. That's why it is async and has to be assumed truly async

> By the way, is the threadpool also involved when doing Task.Yield? 

The current scheduler is involved which can be either a dispatcher aware scheduler or the threadpool

janovesk:
Closing this. Will add a sample how to do burst sends instead.

PR #34, opened 2017-08-21
Add common props and target

andreasohlund:
Just some minor cleanup based on what we learned in https://github.com/Particular/DotNetCoreLaunch/issues/7

### DIFF
--- a/src/Directory.Build.targets
+++ b/src/Directory.Build.targets
@@ -1,0 +1,22 @@
+<Project>
+
+  <PropertyGroup>
+    <Authors>NServiceBus Ltd</Authors>
+    <Company>NServiceBus Ltd</Company>
+    <PackageLicenseUrl>http://particular.net/LicenseAgreement</PackageLicenseUrl>
+    <PackageRequireLicenseAcceptance>true</PackageRequireLicenseAcceptance>
+    <Copyright>Copyright 2010-$([System.DateTime]::UtcNow.ToString(yyyy)) NServiceBus. All rights reserved</Copyright>
+    <PackageTags>nservicebus servicebus msmq cqrs publish subscribe</PackageTags>
+    <PackageIconUrl>http://s3.amazonaws.com/nuget.images/NServiceBus_32.png</PackageIconUrl>
+    <PackageProjectUrl>https://docs.particular.net/nuget/$(PackageId)</PackageProjectUrl>
+    <PackageOutputPath>..\..\nugets</PackageOutputPath>
+    <TargetsForTfmSpecificContentInPackage>$(TargetsForTfmSpecificContentInPackage);IncludePDBsInPackage</TargetsForTfmSpecificContentInPackage>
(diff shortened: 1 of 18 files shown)


andreasohlund:
@timbussmann starting to wonder if we actually need all this given that msmq will never build for netcore? (perhaps we only need a subset)

timbussmann:
well, I think it's easier to just stay to the common approach than some customized config? Nothing on these changes seems to be tied to .net core?

andreasohlund:
Was thinking that most of the workarounds where releated to GitVersion + tests building/running on netcore but I agree that a common approach has merits.

@Particular/msmq-transport-maintainers please review

src/Directory.Build.targets
```diff
@@ -0,0 +1,22 @@
+<Project>
+
+  <PropertyGroup>
+    <Authors>NServiceBus Ltd</Authors>
+    <Company>NServiceBus Ltd</Company>
+    <PackageLicenseUrl>http://particular.net/LicenseAgreement</PackageLicenseUrl>
+    <PackageRequireLicenseAcceptance>true</PackageRequireLicenseAcceptance>
+    <Copyright>Copyright 2010-$([System.DateTime]::UtcNow.ToString(yyyy)) NServiceBus. All rights reserved</Copyright>
+    <PackageTags>nservicebus servicebus msmq cqrs publish subscribe</PackageTags>
```
ramonsmits: Do we need the tags 'servicebus msmq cqrs publish subscribe' in there? Shouldn't that be 'nservicebus transport msmq'?

src/NServiceBus.Transport.Msmq.Tests/MsmqQueueCreatorTests.cs
```diff
@@ -165,12 +165,9 @@ public void Should_not_add_everyone_and_anonymous_to_already_existing_queues()
 
             var existingQueue = GetQueue(testQueueNameForReceiving);
 
-            MessageQueueAccessRights? nullBecauseRevoked;
-            AccessControlEntryType? accessControlEntryType;
 
-            Assert.False(existingQueue.TryGetPermissions(LocalEveryoneGroupName, out nullBecauseRevoked, out accessControlEntryType));
-            Assert.False(existingQueue.TryGetPermissions(LocalAnonymousLogonName, out nullBecauseRevoked, out accessControlEntryType));
-            Assert.IsNull(accessControlEntryType);
+            Assert.False(existingQueue.TryGetPermissions(LocalEveryoneGroupName, out MessageQueueAccessRights? _, out AccessControlEntryType? _));
```
ramonsmits: Nice! Didn't know this is now possible :-)

src/Directory.Build.targets
andreasohlund: Updated

indualagarsamy: I don't see anything wrong with the tags. I think these are just generic tags.

PR #45, opened 2017-10-03
Adding Queue creation PowerShell scripts to the nuget package

indualagarsamy:
1. Installing the nuget package will create a folder called scripts where both the createqueues and deletequeues scripts are copied
2. An API to explicitly not create queues on install has been included.
3. Upon startup if the queue is not found, the exception now provides more details to help the user take the right action
4. Added an unit test to check the not creating the queues functionality.

Documentation:
- https://github.com/Particular/docs.particular.net/pull/3306

### DIFF
--- a/src/NServiceBus.Transport.Msmq.Tests/MsmqQueueCreatorTests.cs
+++ b/src/NServiceBus.Transport.Msmq.Tests/MsmqQueueCreatorTests.cs
@@ -42,20 +42,6 @@
         }
 
         [Test]
-        public void Should_not_create_queue_when_configured_to_not_create_queues_on_install()
-        {
-            var queueName = $"{testQueueNameForReceiving}";
-            var creator = new MsmqQueueCreator(true);
-            var bindings = new QueueBindings();
-
-            bindings.BindSending(queueName);
-
-            creator.CreateQueueIfNecessary(bindings, WindowsIdentity.GetCurrent().Name);
-
-            Assert.False(QueueExists(testQueueNameForReceiving));
-        }
-
-        [Test]
         public void Should_not_create_queue_when_a_remote_queue_is_provided()
         {
             var remoteQueueName = $"{testQueueNameForReceiving}@some-machine";
--- a/src/NServiceBus.Transport.Msmq.Tests/MsmqQueueCreatorTests.cs
+++ b/src/NServiceBus.Transport.Msmq.Tests/MsmqQueueCreatorTests.cs
@@ -45,7 +45,7 @@
         public void Should_not_create_queue_when_configured_to_not_create_queues_on_install()
         {
             var queueName = $"{testQueueNameForReceiving}";
-            var creator = new MsmqQueueCreator(true, true);
+            var creator = new MsmqQueueCreator(true, false);
             var bindings = new QueueBindings();
 
             bindings.BindSending(queueName);
--- a/src/NServiceBus.Transport.Msmq.Tests/MsmqQueueCreatorTests.cs
+++ b/src/NServiceBus.Transport.Msmq.Tests/MsmqQueueCreatorTests.cs
@@ -45,7 +45,7 @@
         public void Should_not_create_queue_when_configured_to_not_create_queues_on_install()
         {
             var queueName = $"{testQueueNameForReceiving}";
-            var creator = new MsmqQueueCreator(true, true);
+            var creator = new MsmqQueueCreator(true);
             var bindings = new QueueBindings();
 
             bindings.BindSending(queueName);
--- a/src/NServiceBus.Transport.Msmq.Tests/MsmqQueueCreatorTests.cs
+++ b/src/NServiceBus.Transport.Msmq.Tests/MsmqQueueCreatorTests.cs
@@ -42,6 +42,20 @@
         }
 
         [Test]
+        public void Should_not_create_queue_when_configured_to_not_create_queues_on_install()
+        {
+            var remoteQueueName = $"{testQueueNameForReceiving}";
+            var creator = new MsmqQueueCreator(true, true);
+            var bindings = new QueueBindings();
+
+            bindings.BindSending(remoteQueueName);
+
+            creator.CreateQueueIfNecessary(bindings, WindowsIdentity.GetCurrent().Name);
+
+            Assert.False(QueueExists(testQueueNameForReceiving));
+        }
+
+        [Test]
         public void Should_not_create_queue_when_a_remote_queue_is_provided()
         {
             var remoteQueueName = $"{testQueueNameForReceiving}@some-machine";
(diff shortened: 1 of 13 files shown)


src/NServiceBus.Transport.Msmq/Scripts/DeleteQueues.ps1
```diff
@@ -0,0 +1,46 @@
+# USAGE:
+
+# To delete the endpoint and all the subqueues, use:
+# DeleteQueuesForEndpoint -EndpointName "myendpoint" 
+
+# To delete a single queue such as Audit or Error, use:
+# DeleteQueue -QueueName -QueueName "error"
```
janovesk: Double `QueueName` here

src/NServiceBus.Transport.Msmq.Tests/MsmqQueueCreatorTests.cs
```diff
@@ -42,6 +42,20 @@ public void Should_create_all_queues()
         }
 
         [Test]
+        public void Should_not_create_queue_when_configured_to_not_create_queues_on_install()
+        {
+            var remoteQueueName = $"{testQueueNameForReceiving}";
```
janovesk: Not a remote queue in this test. Caught you copy-pasting! :)

indualagarsamy:
@janovesk - Thanks for the Ctrl-C, Ctrl-V catch :). All fixed now.

src/NServiceBus.Transport.Msmq/Scripts/CreateQueues.ps1
```diff
@@ -0,0 +1,120 @@
...
+    CreateQueue -QueueName $EndpointName -Account $Account
+
+    # timeout queue
+    CreateQueue -QueueName "$EndpointName.timeouts" -Account $Account
+
+    # timeout dispatcher queue
+    CreateQueue -QueueName "$EndpointName.timeoutsdispatcher" -Account $Account
+
+    # retries queue
+    if ($IncludeRetries) {
```
andreasohlund: No longer needed since the retries satellite is gone in v7

indualagarsamy:
@janovesk @andreasohlund - If I call `DisableInstaller` on the transport, if the msmq subscription storage queue  is configured, it will not be created as well. However, I created a PR to default the subscription storage to use the endpoint name with the "Subscriptions" suffix. In that case, all we need to do is add an extra line to the powershell script to include the subscription queue? We can always have the powershell create the queue to be on the safe side? The user can remove the line or change the queue name? Thoughts?

andreasohlund:
Makes sense to me :+1:

src/NServiceBus.Transport.Msmq/MessagePump.cs
```diff
@@ -202,10 +202,19 @@ bool QueueIsTransactional()
             {
                 return inputQueue.Transactional;
             }
+            catch (MessageQueueException msmqEx)
+            {
+                var error = $"There is a problem with the input inputQueue: {inputQueue.Path}. See the enclosed exception for details.";
+                if (msmqEx.MessageQueueErrorCode == MessageQueueErrorCode.QueueNotFound)
+                {
+                    error = $"The inputQueue: {inputQueue.Path} does not exist. Run the queue creation PowerShell scripts that come packaged with the Nuget package, or explicitly call EnableInstaller() API to create the queues on startup.";
```
andreasohlund: >or explicitly call EnableInstaller() API to create the queues on startup.

how about:

or enable queue creation on startup using `EndpointConfiguration.EnableInstallers()`.

andreasohlund: >Run the queue creation PowerShell scripts that come packaged with the Nuget package

How about:

Run the `/Scripts/CreateQueues.ps1` script included in the `NServiceBus.Transports.Msmq` NuGet package.

indualagarsamy: @andreasohlund - sounds good. I'll incorporate it.

src/NServiceBus.Transport.Msmq/MsmqConfigurationExtensions.cs
```diff
@@ -76,5 +76,21 @@ public static void UseDeadLetterQueueForMessagesWithTimeToBeReceived(this Transp
             Guard.AgainstNull(nameof(config), config);
             config.GetSettings().Set(MsmqTransport.UseDeadLetterQueueForMessagesWithTimeToBeReceived, true);
         }
+
+        /// <summary>
+        /// Disables the automatic queue creation when the endpoint configuration, EnableInstaller is called.
```
andreasohlund: "when installers are enabled using `EndpointConfiguration.EnableInstallers()`." ?

src/NServiceBus.Transport.Msmq/MsmqQueueCreator.cs
```diff
@@ -8,13 +8,18 @@ namespace NServiceBus.Transport.Msmq
 
     class MsmqQueueCreator : ICreateQueues
     {
-        public MsmqQueueCreator(bool useTransactionalQueues)
+        public MsmqQueueCreator(bool useTransactionalQueues, bool disableInstaller = false)
```
andreasohlund: Do we need the default value? 

I think it reads better if we just call the parameter `enabled`? (see below)

src/NServiceBus.Transport.Msmq/MsmqQueueCreator.cs
```diff
@@ -8,13 +8,18 @@ namespace NServiceBus.Transport.Msmq
...
         {
             this.useTransactionalQueues = useTransactionalQueues;
+            this.disableInstaller = disableInstaller;
         }
 
         public Task CreateQueueIfNecessary(QueueBindings queueBindings, string identity)
         {
+            if (disableInstaller)
```
andreasohlund: If we rename like suggested this would be `!enabled` ?

src/NServiceBus.Transport.Msmq/MsmqQueueCreator.cs
indualagarsamy: I used the default value here for the 2nd parameter, otherwise, I'd have to change a whole bunch of existing unit tests to go and add the 2nd parameter.

indualagarsamy:
@andreasohlund - this is good to go. But it needs to be merged after the changes to the default storage queue has been merged. cc / @Particular/msmq-transport-maintainers

src/NServiceBus.Transport.Msmq/MsmqQueueCreator.cs
indualagarsamy: I had to revert my changes. The API is called DisableInstaller and it sets a true. The default value for the 2nd parameter in MsmqQueueCreator cannot also be true. After a "few" failing acceptance tests, I've reverted back, but named the variables better in MsmqQueueCreator. Hope this helps.

andreasohlund:
Think this one needs a rebase now?

indualagarsamy:
Let me rebase.

PR #49, opened 2017-10-05
remove double negative

SimonCropp:


src/NServiceBus.Transport.Msmq.Tests/MsmqQueueCreatorTests.cs
```diff
@@ -45,7 +45,7 @@ public void Should_create_all_queues()
         public void Should_not_create_queue_when_configured_to_not_create_queues_on_install()
         {
             var queueName = $"{testQueueNameForReceiving}";
-            var creator = new MsmqQueueCreator(true, true);
+            var creator = new MsmqQueueCreator(true);
```
indualagarsamy: @SimonCropp - This test isn't valid anymore. As we have kept the MsmqQueueCreator as is. So, its gonna create queues. We can remove this test completely.

PR #52, opened 2017-10-06
Save hot-path allocations by using a local function

MarcinHoppe:
Related to Particular/PlatformDevelopment#1467

### DIFF
--- a/src/NServiceBus.Transport.Msmq/MessagePump.cs
+++ b/src/NServiceBus.Transport.Msmq/MessagePump.cs
@@ -159,11 +159,12 @@ namespace NServiceBus.Transport.Msmq
                     // antecedent task is aborted the continuation will be scheduled. But in this case we don't need to await
                     // the continuation to complete because only really care about the receive operations. The final operation
                     // when shutting down is a clear of the running tasks anyway.
-                    receiveTask.ContinueWith((t, state) =>
+                    void RemoveReceiveTask(Task task, object state)
                     {
-                        var receiveTasks = (ConcurrentDictionary<Task, Task>) state;
-                        receiveTasks.TryRemove(t, out Task _);
-                    }, runningReceiveTasks, TaskContinuationOptions.ExecuteSynchronously)
+                        var receiveTasks = (ConcurrentDictionary<Task, Task>)state;
+                        receiveTasks.TryRemove(task, out Task _);
+                    }
+                    receiveTask.ContinueWith(RemoveReceiveTask, runningReceiveTasks, TaskContinuationOptions.ExecuteSynchronously)
                         .Ignore();
                 }
             }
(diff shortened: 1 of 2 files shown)


src/NServiceBus.Transport.Msmq/MessagePump.cs
```diff
@@ -159,11 +159,12 @@ public void Start(PushRuntimeSettings limitations)
                     // antecedent task is aborted the continuation will be scheduled. But in this case we don't need to await
                     // the continuation to complete because only really care about the receive operations. The final operation
                     // when shutting down is a clear of the running tasks anyway.
-                    receiveTask.ContinueWith((t, state) =>
+                    void RemoveReceiveTask(Task task, object state)
```
danielmarbach: Do we even need the state object when we use a local function? The state object was added to avoid the closure.

MarcinHoppe: Good catch. The local functions becomes a method (I think) and no closure needs to be created.

The fix is on the way.

src/NServiceBus.Transport.Msmq/MessagePump.cs
```diff
@@ -159,12 +159,13 @@ public void Start(PushRuntimeSettings limitations)
                     // antecedent task is aborted the continuation will be scheduled. But in this case we don't need to await
                     // the continuation to complete because only really care about the receive operations. The final operation
                     // when shutting down is a clear of the running tasks anyway.
-                    void RemoveReceiveTask(Task task, object state)
+
+                    void RemoveReceiveTask(Task task)
```
danielmarbach: Can we verify by decompiling it in raw csharp?

bording:
@MarcinHoppe This change actually introduces an allocation, not removed it!

The existing code doesn't isn't closing over variables and state is passed in, so the `Action` needed to call the continuation is allocated once and cached:

image


With the changes introduced in this PR, the local function call has to be wrapped in an `Action` still, and it can no longer be cached:

image

Based on that, this doesn't seem like a change that should be made!

bording:
From what I can see, the main reason this is worse is that the local function is being used as an argument to a method vs. being used directly inside the defining method.

danielmarbach:
@bording thanks, that's what I meant when I said this https://github.com/Particular/NServiceBus.Transport.Msmq/pull/52#discussion_r143173317 changes like the one introduced here need to be verified with decompiling. When I introduced the original version of this code I went through multiple iterations with decompilation.

MarcinHoppe:
@bording @danielmarbach Many thanks for double checking this before I could get it to work. I imagine I could also cache the `Action` wrapper for the local function but that would cancel the benefit.

I am closing this PR because there's no benefit.

PR #54, opened 2017-10-25
New API for what used to be in the MSMQ connection string

indualagarsamy:
Added first level code API for disabling connection caching for sending messages, enabling journaling, overriding the TimeToReachQueue value, disabling dead letter queueing,  and using non-transactional queues instead of supplying these values via a connection string.

If the connection string is passed as shown:
```c#
var transport = endpointConfiguration.UseTransport<MsmqTransport>();
transport.ConnectionString("deadLetter=false;");
```

then the following exception will be thrown prompting to use the newly added code level API.

```c#
var error = @"Passing in MSMQ settings such as DeadLetterQueue, Journaling etc via a connection string is no longer supported.  Use code level API. For example:
To turn off dead letter queuing, use: 
var transport = endpointConfiguration.UseTransport<MsmqTransport>();
transport.DisableDeadLetterQueueing();

To stop caching connections, use: 
var transport = endpointConfiguration.UseTransport<MsmqTransport>();
transport.DisableConnectionCachingForSends();

To use non-transactional queues, use:
var transport = endpointConfiguration.UseTransport<MsmqTransport>();
transport.UseNonTransactionalQueues();

To enable message journaling, use:
var transport = endpointConfiguration.UseTransport<MsmqTransport>();
transport.EnableJournaling();

To override the value of TTRQ, use:
var transport = endpointConfiguration.UseTransport<MsmqTransport>();
transport.TimeToReachQueue(timespanValue);";
```

## Use of Explicit API vs a boolean parameter. 

The use of boolean parameters was considered, for example, instead of "transport.DisableConnectionCachingForSends()" vs "CacheSendConnections(bool value)". 
However these settings that are passed in the connection string are used very sparsely. Also the default for these values are seldom changed. Hence, it made sense to make this API more explicit. For example, queues should be transactional in order to not lose messages. The user should explicitly call, "UseNonTransactionalQueues()" understanding the API and its ramifications. And should only call the API when its needed for the use case warrants. 

## Documentation

https://github.com/Particular/docs.particular.net/pull/3379


indualagarsamy:
@Particular/msmq-transport-maintainers - please review. I think it makes sense to do this now as we are releasing a major as opposed to the next major.

SimonCropp:
should configs take a bool? eg if someone wants to read DoNotCacheConnections from some env variable or config . they will have a bool. so should the API be  `transport.CacheConnections(true/false);`

indualagarsamy:
>should configs take a bool? eg if someone wants to read DoNotCacheConnections from some env variable or config . they will have a bool. so should the API be transport.CacheConnections(true/false);

@SimonCropp - regarding the above. I don't think so. By default, we cache connections and the user shouldn't have to bother with this setting. It is only when they want to explicitly turn off the default behavior. Does that make sense?

src/NServiceBus.Transport.Msmq/MsmqConfigurationExtensions.cs
```diff
@@ -139,5 +139,86 @@ internal static bool GetShouldExecuteInstaller(this SettingsHolder settings)
             }
             return true;
         }
+
+        /// <summary>
+        /// Disables the storing of undeliverable messages in the dead letter queue
```
dvdstelt: Should we not also put a warning here? We do it in the other settings?

src/NServiceBus.Transport.Msmq/MsmqConfigurationExtensions.cs
```diff
@@ -139,5 +139,86 @@ internal static bool GetShouldExecuteInstaller(this SettingsHolder settings)
...
+            config.GetSettings().Set("UseDeadLetterQueue", false);
+        }
+
+        internal static bool GetUseDeadLetterQueue(this SettingsHolder settings)
+        {
+            return settings.GetOrDefault<bool>("UseDeadLetterQueue");
+        }
+
+        /// <summary>
+        /// Does not cache connections.
```
dvdstelt: This comment feels a bit like "Here we enter a loop" just above a "for" loop ;-)
Maybe add something that implies what the difference is?

src/NServiceBus.Transport.Msmq/MsmqConfigurationExtensions.cs
```diff
@@ -139,5 +139,86 @@ internal static bool GetShouldExecuteInstaller(this SettingsHolder settings)
...
+
+        internal static bool GetUseTransactionalQueues(this SettingsHolder settings)
+        {
+            return settings.GetOrDefault<bool>("UseTransactionalQueues");
+        }
+
+        /// <summary>
+        /// Enables the use of journaling messages. Stores a copy of every message received in the journal queue. 
+        /// Should be used ONLY when debugging as it can 
+        /// potentially use up the disk storage quota based on the message volume.
```
dvdstelt: How about "MSMQ disk storage quota"

src/NServiceBus.Transport.Msmq/MsmqConfigurationExtensions.cs
janovesk: Sure

src/NServiceBus.Transport.Msmq/MsmqConfigurationExtensions.cs
janovesk: Added now

src/NServiceBus.Transport.Msmq/MsmqConfigurationExtensions.cs
janovesk: It's actually the MSMQ Journal quota. It's separate from the regular MSMQ quota. 

Added the info

dvdstelt: Oh really? Had no idea. Thanks

janovesk:
Fixed the implementation of defaults. Fixed all the tests and did some wordsmithing on the public API. Also added some startup diagnostics.

Should be good to @indualagarsamy. Have a look and merge if you agree.

andreasohlund:
Seems like TC pulled an old beta, pushed a fix to force beta9

indualagarsamy:
@andreasohlund - looks like the build still failed on the acceptance tests.

andreasohlund:
Found the problem, core issue with transport tests. Will PR on the core and expedite. I propose we move the diagnostics to a seprate PR for now?

andreasohlund:
Pushed a workaround that I think is legit. Will still patch the core though

indualagarsamy:
Discussed with @kbaley and here are a few suggestions for the API. 

```c#
var transport = var transport = endpointConfiguration.UseTransport<MsmqTransport>();
transport.DisableDeadLetterQueue();
transport.DisableConnectionCachingForSends();
transport.UseNonTransactionalQueues();
transport.EnableJournaling();
transport.TimeToReachQueue(timespanValue);";
```

What do you think @Particular/msmq-transport-maintainers ?

andreasohlund:
I like it

indualagarsamy:
Hooray @Particular/msmq-transport-maintainers - build is all green now. Working on the docs and upgrade now.

Issue #55, opened 2017-10-26
Fix the script path included in the NuGet package


indualagarsamy:
Relates to : https://github.com/Particular/docs.particular.net/issues/3365#issuecomment-339246118

andreasohlund:
@indualagarsamy is this done? (PR merged)

indualagarsamy:
Yep. Code and Docs.

PR #56, opened 2017-10-27
Add queue creation/deletion scripts to build output

janovesk:
This will take any script in the **Scripts** folder and add them to the NuGet package. 

A msbuild **.targets** file is also included. It will show the scripts as linked items in a **Scripts** folder in the solution explorer in any project referencing NServiceBus.Transport.Msmq. The content will also be copied to the build output, into a folder called **Scripts**.

Fixes: https://github.com/Particular/NServiceBus.Transport.Msmq/issues/55


src/Directory.Build.targets
```diff
@@ -13,10 +13,21 @@
...
 
+  <ItemGroup>
+    <Content Include="$(ProjectDir)\Scripts\*.targets">     
+    <PackagePath>build</PackagePath>
+    </Content>
+  </ItemGroup>
+
+  <ItemGroup>
+    <Content Include="$(ProjectDir)\Scripts\*.ps1">     
+    <PackagePath>build/Scripts</PackagePath>
```
indualagarsamy: @janovesk - Should the path be "Deployment/Msmq"  instead? ping @andreasohlund

src/NServiceBus.Transport.Msmq/Scripts/NServiceBus.Transport.Msmq.targets
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <ItemGroup>
+    <None Include="$(MSBuildThisFileDirectory)\Scripts\*.ps1">
+      <Link>Scripts\%(RecursiveDir)%(Filename)%(Extension)</Link>
```
janovesk: Probably here you want **Deployment/Scripts** or **Deployment/Msmq** @indualagarsamy? This is the location the scripts end when building a project referencing the nuget.

src/Directory.Build.targets
janovesk: You probably mean line 5 in **NServiceBus.Transport.Msmq.targets** instead. This line only says where to put the scripts in the nuget. They should go under **build** to avoid Nuget from copying the scripts into the project referencing it.

src/NServiceBus.Transport.Msmq/Scripts/NServiceBus.Transport.Msmq.targets
indualagarsamy: yep.

indualagarsamy: I like `Deployment/Scripts` but @andreasohlund mentioned `Deployment/Msmq` in the past. I'll go with whatever you pick. :)

janovesk:
The sql persistence is creating a folder called `NServiceBus.Persistence.Sql\MsSqlServer` in the output directory

janovesk:
How about `Deployment\NServiceBus.Transport.Msmq`? Then we can ask the sql team to do the same. Everything goes under `Deployment\<packagename>`?

andreasohlund:
I like that convention

andreasohlund:
Come to think of it, perhaps ”Deployment” clashes with some other tool or
the users own stuff? Wonder if we should go with the same convention as
sqlp for now?

src/Directory.Build.targets
```diff
@@ -13,10 +13,21 @@
     <TargetsForTfmSpecificContentInPackage>$(TargetsForTfmSpecificContentInPackage);IncludePDBsInPackage</TargetsForTfmSpecificContentInPackage>
   </PropertyGroup>
 
+  <ItemGroup>
```
bording: FYI in my followup PR to move to Particular.Packaging, the need for this Directory.Build.targets file is going away, so these new items would be better in the main csproj instead. 

However, I don't mind doing that as part of my PR, since they are fine where they are for this one.

src/NServiceBus.Transport.Msmq/Scripts/NServiceBus.Transport.Msmq.targets
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
```
bording: This new file should be a props file, not a targets file.

src/NServiceBus.Transport.Msmq/Scripts/NServiceBus.Transport.Msmq.targets
```diff
@@ -0,0 +1,9 @@
...
+<Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <ItemGroup>
```
bording: All imported targets/props files should have

```
 <PropertyGroup>
    <MSBuildAllProjects>$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
  </PropertyGroup>
```

To ensure incremental builds work correctly.

src/NServiceBus.Transport.Msmq/Scripts/NServiceBus.Transport.Msmq.targets
```diff
@@ -0,0 +1,9 @@
...
+    <None Include="$(MSBuildThisFileDirectory)\Scripts\*.ps1">
+      <Link>Scripts\%(RecursiveDir)%(Filename)%(Extension)</Link>
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
```
bording: If this package is used with an SDK-style project, these scripts are going to show up in Solution Explorer, but they won't for either of the old csproj combos (packages.config or package references)

I recommend adding `<Visible>false</Visible>` so that they are hidden everywhere, especially since if someone does try to edit it, they will be modifying the version in the global NuGet package folder.

janovesk: Nice! I didn't know that existed. Not showing is better, totally agree.

src/NServiceBus.Transport.Msmq/Scripts/NServiceBus.Transport.Msmq.targets
janovesk: Check

src/NServiceBus.Transport.Msmq/Scripts/NServiceBus.Transport.Msmq.targets
janovesk: Check

src/NServiceBus.Transport.Msmq/Scripts/NServiceBus.Transport.Msmq.targets
janovesk: Add visibility false and the copy magic to avoid having the linked files/folder show up in solution explorer.

bording:
@indualagarsamy As discussed, I went ahead and pushed the additional changes here that I had originally intended to open a new PR for.

PR #63, opened 2018-05-08
Prepare RTM

bording:


andreasohlund:
Seems to be a few more to fix still, I can look at this tomorrow if needed

PR #67, opened 2018-08-31
Fix nuget warns on master

andreasohlund:


andreasohlund:
@Particular/msmq-transport-maintainers are you happy with this approach to patch master but not release anything?

dvdstelt:
Absolutely

andreasohlund:
@timbussmann I think its ok to squash merge this one to avoid GVT to rev the version?

timbussmann:
we definitely did so before although I'm not sure whether this is really allowed by Gitflow.

PR #78, opened 2019-03-27
Hotfix 1.0.1

andreasohlund:


andreasohlund:
Turns out that this approach doesn't work, will change the PR the adjust the timeout value instead

PR #79, opened 2019-03-27
Make message enumerator timeout configurable

andreasohlund:


andreasohlund:
@ramonsmits I guess you still want the log changes included as well?

andreasohlund:
Note that this PR changes the default to `500ms`, are we happy with that?

danielmarbach:
@andreasohlund I don't know. It is already 50 times better than before so I guess why not. I'm wondering more if we should override this value for the acceptance tests and set it there to 10 milliseconds to speed up the test execution again.

andreasohlund:
Good point, thoughts on that @ramonsmits ?

andreasohlund:
@danielmarbach I've made the tests use `10ms` as the timeout

src/NServiceBus.Transport.Msmq.AcceptanceTests/ConfigureEndpointMsmqTransport.cs
```diff
@@ -13,10 +13,11 @@ public class ConfigureEndpointMsmqTransport : IConfigureEndpointTestExecution
     public Task Configure(string endpointName, EndpointConfiguration configuration, RunSettings settings, PublisherMetadata publisherMetadata)
     {
         queueBindings = configuration.GetSettings().Get<QueueBindings>();
-     
+
         var transportConfig = configuration.UseTransport<MsmqTransport>();
         transportConfig.DisableConnectionCachingForSends();
-        
+        configuration.GetSettings().Set("MessageEnumeratorTimeout", TimeSpan.FromMilliseconds(10));
```
danielmarbach: Do we need the same for the transport tests?

andreasohlund: yea, fixed

andreasohlund:
image

We're now back on par with develop when it comes to how long the ATT's take 👍

src/NServiceBus.Transport.Msmq/MsmqSettings.cs
```diff
@@ -50,6 +51,11 @@ public MsmqSettings(ReadOnlySettings settings)
                 TimeToReachQueue = timeToReachQueue;
             }
 
+            if (settings.TryGet<TimeSpan>("MessageEnumeratorTimeout", out var messageEnumeratorTimeout))
```
danielmarbach: One last thought. I think we should do something like NserviceBus.Transport.Msmq.Key to make it clear it belongs to msmq

andreasohlund:
:+1: fixed now

danielmarbach:
You forgot to change the other keys in ATTs and transport tests

src/NServiceBus.Transport.Msmq.AcceptanceTests/ConfigureEndpointMsmqTransport.cs
danielmarbach: Here

andreasohlund:
LOL, forgot to hit save in VS :facepalm:

PR #80, opened 2019-03-27
Hotfix 1.0.1

### LINKED ISSUE #77: MSMQ results in unnecessary IO when reading from empty queues due to very short timeout value
When reading from queues we are using `MessageQueue.GetMessageEnumerator2().MoveNext(TimeSpan)` in the [message pump](https://github.com/Particular/NServiceBus.Transport.Msmq/blob/0e14e4522329533c262e08e13ded07714d69cdff/src/NServiceBus.Transport.Msmq/MessagePump.cs#L121-L130) in combination with `MessageQueue.Receive(TimeSpan,...)` in the [receive strategy](https://github.com/Particular/NServiceBus.Transport.Msmq/blob/develop/src/NServiceBus.Transport.Msmq/ReceiveStrategy.cs#L30).

The usage of these API's isn't wrong but the supplied timeout value for `MoveNext`  is currently hard coded to 10 milliseconds. It seems that every empty queue to be calling this API 100 times per second with no back off strategy.

**Customer Reported Issue**
A customer who has a large number of endpoints deployed states that the MSMQLOG.BIN file grows very fast. After contacting Microsoft support it shows a very large amount of timeouts are occurring every second causing an issue with overloading the log files. The cause is very likely this very short timeout value.

**Temporary Workaround**
use the following command to check if the trace is started.  (Returns zero if started, non-zero if stopped)
logman query MSMQ -ets

To stop the trace:
logman stop MSMQ –ets

andreasohlund:
Fixes #77 

- Increasing the default value from 10ms to 1 second. 
- Adds a new configuration setting called `NServiceBus.Transport.Msmq.MessageEnumeratorTimeout`,  that can be set to adjust this value if needed. 
- Increasing this value means that the amount of time to see if there are messages in the queue is increased, so this will approximately the same amount of time that the endpoint will wait before it can be shut down. So, for example a 5second value might delay the shutting down of the endpoint by approximately 3s to 5s.  

Note that this PR should not be squashed

### DIFF
--- a/src/NServiceBus.Transport.Msmq.AcceptanceTests/ConfigureEndpointMsmqTransport.cs
+++ b/src/NServiceBus.Transport.Msmq.AcceptanceTests/ConfigureEndpointMsmqTransport.cs
@@ -13,10 +13,11 @@ public class ConfigureEndpointMsmqTransport : IConfigureEndpointTestExecution
     public Task Configure(string endpointName, EndpointConfiguration configuration, RunSettings settings, PublisherMetadata publisherMetadata)
     {
         queueBindings = configuration.GetSettings().Get<QueueBindings>();
-     
+
         var transportConfig = configuration.UseTransport<MsmqTransport>();
         transportConfig.DisableConnectionCachingForSends();
-        
+        configuration.GetSettings().Set("NServiceBus.Transport.Msmq.MessageEnumeratorTimeout", TimeSpan.FromMilliseconds(10));
+
         var routingConfig = transportConfig.Routing();
 
         foreach (var publisher in publisherMetadata.Publishers)
--- a/src/NServiceBus.Transport.Msmq.TransportTests/ConfigureMsmqTransportInfrastructure.cs
+++ b/src/NServiceBus.Transport.Msmq.TransportTests/ConfigureMsmqTransportInfrastructure.cs
@@ -14,6 +14,7 @@ class ConfigureMsmqTransportInfrastructure : IConfigureTransportInfrastructure
     {
         var msmqTransportDefinition = new MsmqTransport();
         settingsHolder = settings;
+        settingsHolder.Set("NServiceBus.Transport.Msmq.MessageEnumeratorTimeout", TimeSpan.FromMilliseconds(10));
         return new TransportConfigurationResult
         {
             TransportInfrastructure = msmqTransportDefinition.Initialize(settingsHolder, null),
--- a/src/NServiceBus.Transport.Msmq/MsmqSettings.cs
+++ b/src/NServiceBus.Transport.Msmq/MsmqSettings.cs
@@ -19,6 +19,7 @@ namespace NServiceBus.Transport.Msmq
             TimeToReachQueue = Message.InfiniteTimeout;
             ScopeOptions = new MsmqScopeOptions();
             LabelGenerator = (headers => string.Empty);
+            MessageEnumeratorTimeout = TimeSpan.FromSeconds(1); //with a 1s timeout a graceful shutdown will take on average 500ms which is acceptable
 
             if (settings == null)
             {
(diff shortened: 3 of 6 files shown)


src/NServiceBus.Transport.Msmq.TransportTests/ConfigureMsmqTransportInfrastructure.cs
```diff
@@ -14,6 +14,7 @@ public TransportConfigurationResult Configure(SettingsHolder settings, Transport
     {
         var msmqTransportDefinition = new MsmqTransport();
         settingsHolder = settings;
+        settingsHolder.Set("NServiceBus.Transport.Msmq.MessageEnumeratorTimeout", TimeSpan.FromMilliseconds(10));
```
ramonsmits: SHould pass `MsmqSettings.MessageEnumeratorTimeout` instead of the magic number.

src/NServiceBus.Transport.Msmq.AcceptanceTests/ConfigureEndpointMsmqTransport.cs
```diff
@@ -13,10 +13,11 @@ public class ConfigureEndpointMsmqTransport : IConfigureEndpointTestExecution
     public Task Configure(string endpointName, EndpointConfiguration configuration, RunSettings settings, PublisherMetadata publisherMetadata)
     {
         queueBindings = configuration.GetSettings().Get<QueueBindings>();
-     
+
         var transportConfig = configuration.UseTransport<MsmqTransport>();
         transportConfig.DisableConnectionCachingForSends();
-        
+        configuration.GetSettings().Set("NServiceBus.Transport.Msmq.MessageEnumeratorTimeout", TimeSpan.FromMilliseconds(10));
```
ramonsmits: Pass `MsmqSettings.MessageEnumeratorTimeout`

ramonsmits: SHouldn't "NServiceBus.Transport.Msmq.MessageEnumeratorTimeout" be a shared string on all locations reading/writing the setting?

andreasohlund: I didn’t want to add any public api hence the magical string for now?

src/NServiceBus.Transport.Msmq.TransportTests/ConfigureMsmqTransportInfrastructure.cs
andreasohlund: Not following? (We want to use 10ms to get fast shutdown)

src/NServiceBus.Transport.Msmq/MsmqSettings.cs
```diff
@@ -50,6 +51,11 @@ public MsmqSettings(ReadOnlySettings settings)
                 TimeToReachQueue = timeToReachQueue;
             }
 
+            if (settings.TryGet<TimeSpan>("NServiceBus.Transport.Msmq.MessageEnumeratorTimeout", out var messageEnumeratorTimeout))
```
ramonsmits: Why is this prefixed with `NServiceBus.Transport.Msmq.` compared to the other settings?

ramonsmits:
@andreasohlund Looking at the AT's it seems it needs to be configured via:

```c#
var messageEnumeratorTimeout = TimeSpan.FromSeconds(10);
var cfg = new EndpointConfiguration("MyEndpoint");
cfg.GetSettings().Set("NServiceBus.Transport.Msmq.MessageEnumeratorTimeout", messageEnumeratorTimeout);
```

Correct?

src/NServiceBus.Transport.Msmq/MsmqSettings.cs
andreasohlund: To make sure it’s unique, we should change the others in the long term

andreasohlund:
Correct

PR #81, opened 2019-03-28
Merge back of 1.0.1

andreasohlund:


andreasohlund:
@bording can you verify this one, GH complain about not being up to date and then added https://github.com/Particular/NServiceBus.Transport.Msmq/pull/81/commits/8e52ce6bf1d4402ab5cb92f078bd32e5e052a9a5 ??

bording:
As mentioned elsewhere, the "update branch" button is never really something you want to use because that will merge the target branch into the PR branch. 

I wish that button would have an option to rebase the PR branch instead, which is what I'd actually want to see happen if updating the branch is important.

PR #82, opened 2019-04-11
Adjust to new recoverability seam

andreasohlund:


SeanFeldman:
TC cannot handle the workaround for CSPROJ @andreasohlund

src/NServiceBus.Transport.Msmq/ReceiveStrategy.cs
```diff
@@ -123,7 +123,7 @@ protected async Task<ErrorHandleResult> HandleError(Message message, Dictionary<
             }
             catch (Exception ex)
             {
-                criticalError.Raise($"Failed to execute recoverability actions for message `{message.Id}`", ex);
+                criticalError.Raise($"Failed to execute recoverability policy for message with native ID: `{message.Id}`", ex);
```
SeanFeldman: Is it **always** native ID though?

andreasohlund: Yes it should be https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Transports/MessageContext.cs#L43

The transport can choose to reuse the nsb id but it would still be the "native id that the transport choose to use"

Makes sense?

andreasohlund:
Ah, so testing locally and then release a core beta when we have checked all transport is the way to go then I guess?

SeanFeldman:
That's what it looks right now. Unless we can do something about TC.

andreasohlund:
Updated to core 7.2.0 beta , this one is ready to go

src/NServiceBus.Transport.Msmq.TransportTests/NServiceBus.Transport.Msmq.TransportTests.csproj
```diff
@@ -14,7 +14,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.9.0" />
-    <PackageReference Include="NServiceBus.TransportTests.Sources" Version="7.1.0" />
+    <PackageReference Include="NServiceBus.TransportTests.Sources" Version="7.2.0-beta0079"/>
```
bording: nitpick: Missing space between quote and the closing tag

andreasohlund: Good catch, fixed!

PR #83, opened 2019-08-02
Prototype with System.Messaging on netstandard2.0 (dotnet core on windows)

valeriob:
This pull request is an experiment to enables using msmq on dotnet core on windows.
I took the source from System.Messaging from here https://github.com/krazure/Experimental.System.Messaging and i had to remove some unsupported features :
-  System.Transactions (lowering the transaction level supported)
image
- MessageQueuePermission.Demand() because i have no clue on how to implement it 

Is it something that can be considered ?


DavidBoike:
Hi @valeriob,

Unfortunately I think it's highly unlikely we'd be able to accept this contribution. First, if we were going to do this, the use of the experimental netcore port would need to be ifdef'ed so that the .NET Framework version (which we wouldn't be able to bump to 472) would continue to run on the .NET Framework System.Messaging APIs, so that DTC could be preserved and we wouldn't cause any breaking changes to our many customers currently using MSMQ on .NET Framework today.

But even then, we'd be taking on the burden of supporting the .NET port of System.Messaging without really any tests to back that up, and that's probably not something we can do.

Unfortunately, the most likely scenario for us to really be able to support MSMQ on .NET Core is for Microsoft to release an official NuGet package for it.

DavidBoike:
Of course we continue to track [this issue](https://github.com/dotnet/corefx/issues/6204) but I see you've already found it and commented. :wink:

valeriob:
Thanks @DavidBoike, i was not expecting for it to be accepted, just to check your position on the topic :D I was not sure if you prefer to keep 1 package or create a new one like NServiceBus.Transport.Msmq.NetCore to underline the breaking change, i think it would be wiser to do so.
What i probably underestimated is the burden to maintain System.Messaging. 
Let's hope someone from Microsoft answer to the topic https://github.com/dotnet/corefx/issues/6204

Issue #68, opened 2018-11-05
Data corruption: At TransactionLevel.None header manipulations done while processing a message are visible in recoverability policy


SzymonPobiega:
When a header is added/deleted/modified in an attempt to process a message and that attempt fails, these header manipulations are visible when the recoverability policy is executed to determine if/how the message should be retried.

kbaley:
Fixed in version 1.0.2

PR #86, opened 2019-10-17
Stop leaking headers

andreasohlund:
Only copies headers when things fail so this should not impact performance for successful processing

Aligns the transport with the behaviour expected by https://github.com/Particular/NServiceBus/pull/5353

Fixes https://github.com/Particular/NServiceBus.Transport.Msmq/issues/68

### DIFF
--- a/src/NServiceBus.Transport.Msmq/NoTransactionStrategy.cs
+++ b/src/NServiceBus.Transport.Msmq/NoTransactionStrategy.cs
@@ -32,7 +32,7 @@ namespace NServiceBus.Transport.Msmq
                 {
                     message.BodyStream.Position = 0;
 
-                    await HandleError(message, headers, exception, transportTransaction, 1).ConfigureAwait(false);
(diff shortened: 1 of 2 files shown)


andreasohlund:
I think it might be easier just to fix the header leak on master instead of removing the test?

danielmarbach:
@andreasohlund I was trying to wrap my head around why we need to reextract in on transaction strategy and then always copy in the MoveToError. So I sent in an alternate proposal that would move the responsibility to reextract the headers again into the MoveMethod to make it consistent. So there is no other way anymore but using the original headers on the message. Not sure what the perf implications are though but given that is on the slow exception path anyway I'm not sure we have to worry about that too much

https://github.com/Particular/NServiceBus.Transport.Msmq/pull/87

andreasohlund:
Closed in favour of https://github.com/Particular/NServiceBus.Transport.Msmq/pull/87

PR #87, opened 2019-10-18
Fixheaderleak alternative

danielmarbach:
Alternative to #86


andreasohlund:
I like this, much cleaner to always re-extract them from the "source" in one place. Since this happens only during failures I'd say perf implications is acceptable

PR #85, opened 2019-10-17
Update NServiceBus packages

timbussmann:

PR #88, opened 2019-10-18
Ignore problematic commit that bumps to 7.2

andreasohlund:
I was obviously wrong 🤦‍♂ 

https://github.com/Particular/NServiceBus.Transport.Msmq/pull/85#issuecomment-543632052


SeanFeldman:
Happens 🙂

PR #110, opened 2020-03-02
Update test projects to NServiceBus 7.2.3

bording:
For your review @mauroservienti

### DIFF
--- a/src/NServiceBus.Transport.Msmq.AcceptanceTests/NServiceBus.Transport.Msmq.AcceptanceTests.csproj
+++ b/src/NServiceBus.Transport.Msmq.AcceptanceTests/NServiceBus.Transport.Msmq.AcceptanceTests.csproj
@@ -13,10 +13,10 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.9.0" />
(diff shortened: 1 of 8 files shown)


mauroservienti:
@bording this is failing because a 7.2.3 version of the `NServiceBus.Testing.Fakes.Sources` doesn't exist, on nuget the latest is 7.1.12

bording:
The test failures here are due to the fact that we have fixed the header leak bug on master in #87, but we have not backported and released it yet. It's still a problem on `release-1.0` and for the latest released version of the transport.

andreasohlund:
Is this still relevant?

bording:
I believe so, yes. This is still something that demonstrates that the released version of the transport fails tests.

ramonsmits:
@bording @andreasohlund why is this onhold? There is currently an enhancement release and we discovered another bug. We can easily resolve this PR, including the bug discovered and release 1.0.2.

bording:
It's on hold because the bug fix on `master` is not currently on the release branch that this PR is targeting, so the tests that 7.2.3 bring in fail.

mikeminutillo:
Is the core purpose of this PR updating the test projects or is it the fix to header handling? Especially for release notes purposes, I think it's the latter. In which case, the name should be something like **Prevent modification of headers before sending to error queue**. Or, as this is a bug fix, should there be an issue named **Changes to headers during failed message processing may end up in the error queue** that gets associated with the milestone/release notes?

PR #125, opened 2020-05-13
Dynatrace causes header corruption which results in failed 

ramonsmits:
Based on user report Dynatrace can add binary data to the MSMQ Message extension which already contains data serialized by NServiceBus. By ignoring everything after the last `</ArrayOfHeaderInfo>` tag this binary data does not affect header deserialization and preventing messages to be treated as POISON messages.

```
System.InvalidOperationException: There is an error in XML document (31, 21). ---> System.Xml.XmlException: '.', hexadecimal value 0x00, is an invalid character. Line 31, position 21.
```

Existing code was already stripping trailing null's and was doing an inefficient byte array to string to Stream conversion. This change does "last index" but after that efficiently creates a MemoryStream preventing the string creation.

This change is to apply the Robustness principle and be liberal in what to accept.

### DIFF
--- a/src/NServiceBus.Transport.Msmq.Tests/MsmqUtilitiesTests.cs
+++ b/src/NServiceBus.Transport.Msmq.Tests/MsmqUtilitiesTests.cs
@@ -138,12 +138,10 @@
         {
             var expected = "Hello World";
 
-            Console.Out.WriteLine(sizeof(char));
             var message = MsmqUtilities.Convert(new OutgoingMessage("message id", new Dictionary<string, string>
             {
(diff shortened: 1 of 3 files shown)


ramonsmits:
FYI: This fix is available as a pre-release

https://www.myget.org/feed/particular/package/nuget/NServiceBus.Transport.Msmq/1.1.0-alpha.97

```ps1
PM> Install-Package NServiceBus.Transport.Msmq -Version 1.1.0-alpha.97 -Source https://www.myget.org/F/particular/api/v3/index.json
```

Issue #1, opened 2017-05-12
Support "soft" TTBR


andreasohlund:
Since the current solution is practically unusable given that we have audit on by default we should consider a soft TTBR similar to sqlserver that we can fallback to.

Requested by user  -https://github.com/Particular/NServiceBus/issues/4698#issuecomment-301070706

PR #138, opened 2020-06-11
Implement Soft TTBR option

mikeminutillo:
Resolves https://github.com/Particular/NServiceBus.Transport.Msmq/issues/1

Provides an API on the transport settings that bypasses the MSMQ Native Time to be Received and relies on NSB Headers to do TTBR instead. When this is set, messages with an expired TTBR will not be discarded until they are read by a receiving endpoint.

```csharp
var transport = endpointConfiguration.UseTransport<MsmqTransport>();
transport.DisableNativeTimeToBeReceived();
```


src/NServiceBus.Transport.Msmq/ReceiveStrategy.cs
```diff
@@ -101,6 +102,11 @@ protected void MovePoisonMessageToErrorQueue(Message message, MessageQueueTransa
 
         protected async Task<bool> TryProcessMessage(string messageId, Dictionary<string, string> headers, Stream bodyStream, TransportTransaction transaction)
         {
+            if(ttbrStrategy.DiscardDueToElapsedTtbr(headers))
```
ramonsmits, line 105: @mikeminutillo Does this truly means it is discarded? Meaning, the message will not be send to the audit queue?

mikeminutillo: Correct. We never enter the pipeline so the message is never audited.

ramonsmits:
Reasons for automatically only using soft TTBR:

- Convenience
- If at a later stage user could use outbox of transaction mode ReceiveOnly, we automatically would use native TTBR again

Reasons for explicitly disabled native TTBR:

- Users making the false assumption that native TTBR is active and expired message to be automatically purged by MSMQ and could cause outage.
- If user explicitly disabled native TTBR and in the future use Outbox, then there is not reason to have native TTBR disabled.

ramonsmits:
Closing in favor of https://github.com/Particular/NServiceBus.Transport.Msmq/pull/145

PR #143, opened 2020-06-11
Support reading instance mapping webservers by supporting Uri's

### LINKED ISSUE #135: Support URIs for the instance-mapping file
file based routing supports URIs instead of simple file paths to load the routing file from version 1.1 (http and file scheme only). When using MSMQ it's likely that you will use the routing file together with the instance mapping file. But the instance mapping file can't be hosted on a webserver.

I think we should align those APIs capabilities as they are intended to be used together and add URI support for the instance mapping.

ramonsmits:
Fixes #135


ramonsmits:
Simple solution but not certain. It probably requires some refactoring which is why I didn't' add tests yet. It is not documented how `XmlReader.Create` does file locks. Maybe the code can just be simplified to use the string overload direction on all cases.

src/NServiceBus.Transport.Msmq/InstanceMapping/InstanceMappingFileAccess.cs
```diff
@@ -8,10 +9,20 @@ class InstanceMappingFileAccess : IInstanceMappingFileAccess
     {
         public XDocument Load(string path)
         {
-            using (var file = new FileStream(path, FileMode.Open, FileAccess.Read, FileShare.ReadWrite))
-            using (var reader = XmlReader.Create(file))
+            if (Uri.TryCreate(path, UriKind.Absolute, out var uri) && !uri.IsFile)
```
mikeminutillo, line 12: Is it worth getting rid of the `!` and reversing the order of the sections? i.e.

```csharp
if (Uri.TryCreate(path, UriKind.Absolute, out var uri) && uri.IsFile)
{
  // Manually create the FileStream
}
else
{
  // Just pass what we have to XmlReader.Create
}
```

If we cannot create a Uri currently we try and do FileStream path which seems more narrow than the XDocument Load.

mikeminutillo:
It may be worth pushing `Uri` all the way out to the configuration api. [That's how FileBasedRouting does it](https://github.com/ParticularLabs/NServiceBus.FileBasedRouting/blob/develop/src/NServiceBus.FileBasedRouting/FileBasedRoutingConfigExtensions.cs#L66)

PR #145, opened 2020-06-15
Disable Native TTBR in Transactions

### LINKED ISSUE #1: Support "soft" TTBR

mikeminutillo:
Resolves #1 

When combining a message with TTBR and transactions, currently you get an exception. This PR creates a setting that causes the system to fall back on "soft" TTBR when this happens.

- Discards messages with an expired TTBR header when received by an endpoint (Soft TTBR)
  - This behavior can be turned off with an API: `endpointConfiguration.IgnoreIncomingTimeToBeReceivedHeaders();`
- Adds configuration switch: `endpointConfiguration.DisableNativeTimeToBeReceivedInTransactions();`
  - If set, then messages sent in a transaction skip native TTBR (and fall back on soft TTBR)
  - If not set, then messages sent in a transaction with TTBR throw an exception.

### DIFF
--- a/src/NServiceBus.Transport.Msmq.AcceptanceTests/ConfigureEndpointMsmqTransport.cs
+++ b/src/NServiceBus.Transport.Msmq.AcceptanceTests/ConfigureEndpointMsmqTransport.cs
@@ -17,7 +17,7 @@ public class ConfigureEndpointMsmqTransport : IConfigureEndpointTestExecution
         var transportConfig = configuration.UseTransport<MsmqTransport>();
         transportConfig.DisableConnectionCachingForSends();
         configuration.GetSettings().Set("NServiceBus.Transport.Msmq.MessageEnumeratorTimeout", TimeSpan.FromMilliseconds(10));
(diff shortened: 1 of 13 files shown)


mikeminutillo:
The two tests that fail assert that messages sent to error or audit queues should not honor TTBR. For soft TTBR that means that we need the option to ignore the header on incoming messages.

mikeminutillo:
Added hidden configuration switch to disable the Soft TTBR behavior. This will be needed by ServiceControl to ignore TTBR on incoming audit and error messages.
```csharp
config.GetSettings().Set("IgnoreIncomingTimeToBeReceivedHeaders", true)
```

ramonsmits:
@mikeminutillo If we are already exposing this undocumented setting because our tooling needs it then I would rather just have an API for it. Still, it feels a bit strange. I wouldn't have any problem with just removing the TTBR header from the header collection. From memory I think ASB has some transport headers that are also not audited.

PR #146, opened 2020-06-16
Attempt stricter schema validation on instance mapping XML

### LINKED ISSUE #134: Routing: Prevent mistakes in instance mapping file
It's possible to configure all sorts of attributes in the instance mapping file for an endpoint instance (e.g. `machine`) but in case an user doesn't exactly use the casing as expected, the transport will not be able to find the attribute (unless it explicitly looks for all possible mutations).

At first glance I see two different implementations:
- Make the schema for the instance mapping file more strict to not allow arbitrary attributes and include the `machine` and `queue` attributes. This will cause exceptions if anything else is configured. In case of runtime changes, this might result in changes not being picked up since after startup validation changes will let the endpoint keep the last known valid mapping configuration. As the instance mapping file is currently limited to MSMQ, there is no more need for the support of arbitrary information in the mapping file.
- Make the lookup to the `EndpointInstances.Properties` property more forgiving so that a lookup for `machine` will also return if the attribute was declared `MACHINE`. This easy to implement by passing  `StringComparer.InvariantCultureIgnoreCase` to a newly created Dictionary.

/cc @bording @SzymonPobiega

mikeminutillo:
Resolves #134 

Attempts to validate the instance mapping XML document against a strict schema that does not accept unused attributes.

If this fails then a warning is logged and the instance mapping XML is validated using the old method. The warning is only logged when the validation state changes. i.e. If it keeps falling back, it only logs once. This is to avoid flooding the logs with validation messages.

You can opt-in to strict enforcement by calling:

```csharp
endpointConfiguration.UseTransport<MsmqTransport>()
  .Routing()
  .InstanceMapping()
  .EnforceStrictSchemaValidation();
```

This will cause an exception to be thrown if the instance mapping file does not adhere to the stricter schema that does not accept unused attributes.


ramonsmits:
@mikeminutillo Maybe we can add a form of state change detection? Only log the WARN each time a change is detected instead of every refresh interval?

PR #148, opened 2020-06-16
Ignoring MessageQueue.Create test

ramonsmits:
Ignoring MessageQueue.Create test. Default permission behavior is dependent on if the computer is part of a domain. When part of a domain Everyone and Anonymous both do not get permission to write to the queue.

### DIFF
--- a/src/NServiceBus.Transport.Msmq.Tests/MsmqQueueCreatorTests.cs
+++ b/src/NServiceBus.Transport.Msmq.Tests/MsmqQueueCreatorTests.cs
@@ -112,27 +112,6 @@
             Assert.False(queue.Transactional);
         }
 
-        [Test, Ignore("Default permission behavior is dependent on if the computer is part of a domain. When part of a domain, Everyone and Anonymous both do not get permissions to write to the queue.")]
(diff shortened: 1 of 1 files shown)


kbaley:
Shouldn't we delete the test completely? Or do you want to leave this in as a warning for future people not to write the test again?

mikeminutillo:
I agree with @kbaley. Why are we asserting this at all? We don't want these permissions added by default. We don't have code to do it. It happens as a side-effect of the OS.

If anything, I'd want a test that asserts the opposite. Unfortunately, we don't have a way of making that test pass so there's no value in writing it.

ramonsmits:
@mikeminutillo @kbaley We could split the test, have a test per environment (in a domain, without a domain) as it does test behavior. I'm fine with removing the test but lets check if the queue creation logic has some code comments on assumed permissions.

mikeminutillo:
> it does test behavior.

It doesn't test our behavior though does it? We are not reliant on these permissions getting set up (if we're doing it right).

We do have code to detect this condition and tell the user that it's a problem they should fix.

PR #149, opened 2020-06-18
Simplify the pump stop to no longer use the task tracking approach and wait until all message processing is done

danielmarbach:
Alternative to #150

We have this cruft

Transports have different and inconsistent shutdown behavior when it comes to the time they wait until all pending transport operations are completed.

- RabbitMQ: Indefinitely waits until all pending transport operations are completed
- AmazonSQS: Indefinitely waits until all pending transport operations are completed
- MSMQ: Waits until all pending transport operations are completed for a maximum of thirty seconds and then forcefully shuts down
- SqlTransport: Waits until all pending transport operations are completed for a maximum of thirty seconds and then forcefully shuts down
- Azure Storage Queues: Waits until all pending transport operations are completed for a maximum of thirty seconds and then forcefully shuts down
- Azure Service Bus: Indefinitely waits until all pending transport operations are completed

This PR removes the convoluted task tracking approach and makes the transport wait until all pending transport operations are done. I do believe until we have a proper way to pass in cancellation to the transport this is a sensible and safe default. Shutdown can always be enforced by the host by killing the endpoint if needed.

### DIFF
--- a/src/NServiceBus.Transport.Msmq/MessagePump.cs
+++ b/src/NServiceBus.Transport.Msmq/MessagePump.cs
@@ -1,9 +1,7 @@
 namespace NServiceBus.Transport.Msmq
 {
     using System;
-    using System.Collections.Concurrent;
     using System.Diagnostics;
-    using System.Linq;
     using System.Messaging;
     using System.Threading;
     using System.Threading.Tasks;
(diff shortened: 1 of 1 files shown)


ramonsmits:
Closing in favor of #150

PR #153, opened 2020-06-19
Machine name remote check now checks  Ipv6 loopback and using DNS lookup

### LINKED ISSUE #133: Identifying remote queue using IP instead of machine names
We currently compare machine names to identify if the queue about to be created is in a remote machine or not, see https://github.com/Particular/NServiceBus/blob/33947a34659fbad5e9636873aefe7ab2a6077533/src/NServiceBus.Core/Transports/Msmq/MsmqQueueCreator.cs#L35

Wouldn't it be better to identify if the machine is remote or not via:

``` c#
static bool? IsLocalIpAddress(string hostName)
{
    if (string.IsNullOrWhiteSpace(hostName))
    {
        return null;
    }
    try
    {
        var hostIPs = Dns.GetHostAddresses(hostName);
        var localIPs = Dns.GetHostAddresses(Dns.GetHostName());

        if (hostIPs.Any(hostIp => (IPAddress.IsLoopback(hostIp) || (localIPs.Contains(hostIp)))))
        {
            return true;
        }
    }
    catch
    {
        return null;
    }
    return false;
}
```

ramonsmits:
Fixes #133

The remote check not only checks the machine name but also checks if the address is an Ipv6 loopback address and uses DNS lookup to validate if the queues are remote.

### DIFF
--- a/src/NServiceBus.Transport.Msmq.Tests/MsmqAddressTests.cs
+++ b/src/NServiceBus.Transport.Msmq.Tests/MsmqAddressTests.cs
@@ -43,13 +43,13 @@ namespace NServiceBus.Transport.Msmq.Tests
         [TestCase(".")]
         public void If_machine_is_looplocal_is_specified_is_remote_should_be_false(string machine)
         {
-            Assert.IsFalse(MsmqAddress.Parse("replyToAddress@" + machine).IsRemote);
+            Assert.IsFalse(MsmqAddress.Parse("replyToAddress@" + machine).IsRemote());
         }
 
         [Test]
         public void If_local_machine_name_is_remote_should_be_false()
         {
-            Assert.IsFalse(MsmqAddress.Parse("replyToAddress@" + Environment.MachineName).IsRemote);
+            Assert.IsFalse(MsmqAddress.Parse("replyToAddress@" + Environment.MachineName).IsRemote());
         }
 
         [Test]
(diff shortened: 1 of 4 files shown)


src/NServiceBus.Transport.Msmq.Tests/MsmqAddressTests.cs
```diff
@@ -11,10 +12,10 @@ public class MsmqAddressTests
         public void If_both_addresses_are_specified_via_host_name_it_should_not_convert()
         {
             var address = new MsmqAddress("replyToAddress", "replyToMachine");
-            var returnAddress = address.MakeCompatibleWith(new MsmqAddress("someQueue","destinationmachine"), _ =>
-            {
-                throw new Exception("Should not call the lookup method");
-            });
+            var returnAddress = address.MakeCompatibleWith(new MsmqAddress("someQueue", "destinationmachine"), _ =>
+             {
```
kbaley, line 16: Extra space

src/NServiceBus.Transport.Msmq/MsmqAddress.cs
```diff
@@ -48,14 +49,37 @@ public static MsmqAddress Parse(string address)
             return new MsmqAddress(queue, machineName);
         }
 
-        public bool IsRemote => Machine != RuntimeEnvironment.MachineName;
+        public bool IsRemote => Machine != RuntimeEnvironment.MachineName && IsLocalIpAddress(Machine) != true;
```
mikeminutillo, line 52: How time consuming is this? Should it be lazily evaluated?

ramonsmits: As discussed, the remote check is done only at startup and we do not have to worry about performance impact. I've modified it from property to function to imply that this isn't a cheap operation.

PR #175, opened 2020-09-24
Add Particular.CodeRules 0.7.0

kbaley:
Remove ReSharper artifacts


adamralph (approved):
@kbaley I think we should squash all these commits before merging.

PR #197, opened 2021-01-26
Update transport seam

timbussmann:
WIP


src/NServiceBus.Transport.Msmq/MsmqTransport.cs
```diff
@@ -106,5 +112,170 @@ static void ValidateIfDtcIsAvailable(ReadOnlySettings settings)
...
+        /// <summary>
+        /// Indicates whether queues should be automatically created. Disables the automatic queue creation when installers are enabled using `EndpointConfiguration.EnableInstallers()`.
+        /// </summary>
+        /// <remarks>
+        /// With installers enabled, required queues will be created automatically at startup.While this may be convenient for development,
+        /// we instead recommend that queues are created as part of deployment using the CreateQueues.ps1 script included in the NuGet package.
+        /// The installers might still need to be enabled to fulfill the installation needs of other components, but this method allows
+        /// scripts to be used for queue creation instead.
+        /// </remarks>
+        public bool ExecuteInstaller { get; set; } = true;
```
timbussmann, line 159: I think `CreateQueues` would be more appropriate

src/NServiceBus.Transport.Msmq/MsmqTransport.cs
```diff
@@ -106,5 +112,170 @@ static void ValidateIfDtcIsAvailable(ReadOnlySettings settings)
...
+        /// When set to <c>true</c>, disables native Time-To-Be-Received (TTBR) when combined with transactions.
+        /// </summary>
+        public bool DisableNativeTtbrInTransactions { get; set; } = false;
+
+        //TODO improve naming/docs
+        /// <summary>
+        /// The user account that will be configured with access rights to the queues created by the transport. When not set, the current user will be used.
+        /// This setting is not relevant, if queue creation has been disabled.
+        /// </summary>
+        public string Username { get; set; }
```
timbussmann, line 224: QueueUser? QueueOwner? UserAccount?

src/NServiceBus.Transport.Msmq/DeadLetterQueueOptionExtensions.cs
```diff
@@ -18,8 +19,17 @@ public static class DeadLetterQueueOptionExtensions
         public static void UseDeadLetterQueue(this ExtendableOptions options, bool enable = true)
         {
             Guard.AgainstNull(nameof(options), options);
-            var ext = options.GetExtensions();
-            ext.Set(KeyDeadLetterQueue, enable);
+            options.GetDispatchProperties()[KeyDeadLetterQueue] = enable.ToString();
```
SzymonPobiega, line 22: Nitpicking: do we need to pass invariant culture here, just in case?

src/NServiceBus.Transport.Msmq/JournalOptionExtensions.cs
```diff
@@ -18,8 +19,18 @@ public static class JournalOptionExtensions
         public static void UseJournalQueue(this ExtendableOptions options, bool enable = true)
         {
             Guard.AgainstNull(nameof(options), options);
-            var ext = options.GetExtensions();
-            ext.Set(KeyJournaling, enable);
+
+            options.GetDispatchProperties()[KeyJournaling] = enable.ToString();
```
SzymonPobiega, line 23: Nitpicking: do we need to pass invariant culture here, just in case?

src/NServiceBus.Transport.Msmq/MessagePump.cs
```diff
@@ -9,69 +9,86 @@ namespace NServiceBus.Transport.Msmq
...
+        public Task StartReceive()
         {
             MessageQueue.ClearConnectionCache();
 
-            maxConcurrency = limitations.MaxConcurrency;
-            concurrencyLimiter = new SemaphoreSlim(limitations.MaxConcurrency, limitations.MaxConcurrency);
-            cancellationTokenSource = new CancellationTokenSource();
+            if (receiveSettings.PurgeOnStartup)
+            {
+                inputQueue.Purge();
```
SzymonPobiega, line 79: Why have we moved the purge? Does it change the bahavior in any way?

src/NServiceBus.Transport.Msmq/NonDurable/DetermineMessageDurabilityBehavior.cs
```diff
@@ -0,0 +1,33 @@
+﻿namespace NServiceBus.Transport.Msmq
+{
+    using System;
+    using System.Collections.Concurrent;
+    using System.Threading.Tasks;
+    using Pipeline;
+
+    class DetermineMessageDurabilityBehavior : IBehavior<IOutgoingLogicalMessageContext, IOutgoingLogicalMessageContext>
```
SzymonPobiega, line 8: Why can't we do this in the dispatcher? Is it because we don't know the message type in the transport operation? The lack of message type on the operation created for me some friction in SQS, maybe we should consider adding it? 

I am concerned because this API won't work in raw context. Not that I am a big fan of it but...

src/NServiceBus.Transport.Msmq/NonDurable/ExpressMessageExtensions.cs
```diff
@@ -0,0 +1,22 @@
...
+
+    /// <summary>
+    /// Provides extension methods to configure express messages.
+    /// </summary>
+    public static class ExpressMessageExtensions
+    {
+        /// <summary>
+        /// Sets the function to be used to evaluate whether a type is an express message or not.
+        /// </summary>
+        public static void DefiningExpressMessagesAs(this RoutingSettings<MsmqTransport> routingSettings, Func<Type, bool> expressMessageConvention)
```
SzymonPobiega, line 14: Related to my comment above, maybe we could change the type of the callback here to be of type `Func<TransportMessage, boo>` so that the application of the convention can happen in the transport?

SzymonPobiega (approved):
I am OK with it but have some questions. Neither of them is a blocked so I mark it `approved` but especially the type-based callback concerns me a bit.

src/NServiceBus.Transport.Msmq/DeadLetterQueueOptionExtensions.cs
timbussmann: Good point but I don't think we do, bool's `ToString` implementation is basically hardcoded: https://referencesource.microsoft.com/#mscorlib/system/boolean.cs,6c8fb8c92a24ec37

src/NServiceBus.Transport.Msmq/JournalOptionExtensions.cs
timbussmann: see previous comment

src/NServiceBus.Transport.Msmq/MessagePump.cs
timbussmann: I was thinking it should probably only purge once you start to receive the queue, not when you define the queues already. But thinking a bit more about it, maybe we need to purge earlier. In core, this is actually important as feature startup tasks (before transport initialize & start) could send subscription messages, which would now be purged once you start the receiver.

So in essence, purging would have to go pack to when we initialize the transport, which I dislike a bit but I guess it needs to behave that way for compatibility reasons.

src/NServiceBus.Transport.Msmq/NonDurable/DetermineMessageDurabilityBehavior.cs
timbussmann: >  Is it because we don't know the message type in the transport operation? The lack of message type on the operation created for me some friction in SQS, maybe we should consider adding it?

yes that is the issue. That might be some addition to be added to the seam.

> I am concerned because this API won't work in raw context. 

yes, this is currently a feature that MSMQ provides to Core and isn't available for raw modes.

src/NServiceBus.Transport.Msmq/NonDurable/ExpressMessageExtensions.cs
timbussmann: as you said yourself in the previous comment, this doesn't give us information about the message type unfortunately. Given this functionality is message-type convention based, this is kinda crucial

src/NServiceBus.Transport.Msmq/MessagePump.cs
timbussmann: I moved the purging to Initialize, but due to the current usage of the API by core, this doesn't really change the behavior at all. This requires https://github.com/Particular/NServiceBus/pull/5906 in addition to make the behavior similar to v7

src/NServiceBus.Transport.Msmq.AcceptanceTests/ConfigureEndpointMsmqTransport.cs
```diff
@@ -73,6 +70,17 @@ public Task Cleanup()
 
         return Task.FromResult(0);
     }
+}
+
+class TestableMsmqTransport : MsmqTransport
```
andreasohlund, line 75: Nitpick: Move to separate file?

src/NServiceBus.Transport.Msmq.AcceptanceTests/TimeToBeReceived/When_sending_message_with_ttbr.cs
```diff
@@ -75,7 +75,11 @@ public async Task Suppress_native_ttbr_if_enabled_inside_transaction()
         {
             var context = await Scenario.Define<Context>()
                 .WithEndpoint<Sender>(endpoint => endpoint
-                    .CustomConfig(endpointConfiguration => endpointConfiguration.UseTransport<MsmqTransport>().DisableNativeTimeToBeReceivedInTransactions())
+                    .CustomConfig(endpointConfiguration =>
+                    {
+                        var transportSettings = (MsmqTransport)endpointConfiguration.ConfigureTransport();
+                        transportSettings.DisableNativeTtbrInTransactions = true;
```
andreasohlund, line 81: Should this be renamed to `DisableNativeTTBRInTransactions` ?

src/NServiceBus.Transport.Msmq.AcceptanceTests/When_enabling_journaling.cs
```diff
@@ -22,10 +22,10 @@ public async Task Verify_journaling_flag(bool global)
                 await Scenario.Define<Context>()
                 .WithEndpoint<Endpoint>(b => b.CustomConfig(c =>
                 {
-                    var t = c.UseTransport<MsmqTransport>();
+                    var t = (MsmqTransport)c.ConfigureTransport();
```
andreasohlund, line 25: Would a `c.ConfigureTransport<MsmqTransport>()` be an option to avoid the ugly cast?

src/NServiceBus.Transport.Msmq.AcceptanceTests/When_setting_label_generator.cs
```diff
@@ -76,7 +76,8 @@ public Endpoint()
                 EndpointSetup<DefaultServer>(c =>
                 {
                     c.AuditProcessedMessagesTo("labelAuditQueue");
-                    c.UseTransport<MsmqTransport>().ApplyLabelToMessages(GetMessageLabel);
+                    var transportSettings = (MsmqTransport)c.ConfigureTransport();
+                    transportSettings.ApplyLabel = GetMessageLabel;
```
andreasohlund, line 80: Could be more explicit as `ApplyCustomLabelToOutgoingMessages`

src/NServiceBus.Transport.Msmq/NonDurable/ExpressAttribute.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace NServiceBus.Transport.Msmq
+{
+    using System;
+
+    /// <summary>
+    /// Attribute to indicate that the message should not be written to disk.
+    /// This will make the message vulnerable to server crashes or restarts.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Interface)]
+    public sealed class ExpressAttribute : Attribute
```
andreasohlund, line 10: I though this was removed in v8? 

See https://docs.particular.net/nservicebus/upgrades/7to8/#non-durable-messaging and https://particularsoftware.slack.com/archives/C8A66739A/p1602158814032800 ?

andreasohlund (commented):
Some nitpicks and a question around express message support

src/NServiceBus.Transport.Msmq/NonDurable/ExpressAttribute.cs
timbussmann: hm interesting, why isn't MSMQ mentioned given it does support it?

src/NServiceBus.Transport.Msmq.AcceptanceTests/When_setting_label_generator.cs
timbussmann: yeah 👍

src/NServiceBus.Transport.Msmq.AcceptanceTests/When_enabling_journaling.cs
timbussmann: yes, I'm planning to add that to the testing framework, some other downstreams have used their own extension method to do that.

src/NServiceBus.Transport.Msmq.AcceptanceTests/TimeToBeReceived/When_sending_message_with_ttbr.cs
timbussmann: might make more sense. The style-checks will complain about that I think but we can ignore that I guess

src/NServiceBus.Transport.Msmq.AcceptanceTests/When_enabling_journaling.cs
timbussmann: here you go: https://github.com/Particular/NServiceBus/pull/5910 😁

src/NServiceBus.Transport.Msmq/ReceiveStrategy.cs
```diff
@@ -102,21 +101,17 @@ protected void MovePoisonMessageToErrorQueue(Message message, MessageQueueTransa
...
-                var messageContext = new MessageContext(messageId, headers, body, transaction, tokenSource, new ContextBag());
-
-                await onMessage(messageContext).ConfigureAwait(false);
+            var body = await ReadStream(bodyStream).ConfigureAwait(false);
+            var messageContext = new MessageContext(messageId, headers, body, transaction, new ContextBag());
 
-                return tokenSource.Token.IsCancellationRequested;
-            }
+            await onMessage(messageContext).ConfigureAwait(false);
+            return false; //TODO can this be completely removed or will this be readded with CTS support?
```
timbussmann, line 114: @andreasohlund with the removal of the token from the MessageContext, it looks like this doesn't need to return any bool result at all anymore. Is this intended or are there more changes coming back?

src/NServiceBus.Transport.Msmq.AcceptanceTests/When_setting_label_generator.cs
timbussmann: changed

src/NServiceBus.Transport.Msmq.AcceptanceTests/TimeToBeReceived/When_sending_message_with_ttbr.cs
timbussmann: I renamed this to `UseNonNativeTimeToBeReceivedInTransactions` to be more clear about what this setting does. Is that ok for you too @andreasohlund ?

src/NServiceBus.Transport.Msmq/ReceiveStrategy.cs
andreasohlund: It wasn't mentioned since MSMQ wasn't supposed to be supported in v8. That said MSMQ will only change its behaviour when running with non-transactional queues and that isn't really supported since SC can't handle it (running with non-transactional queues requires all queues in the entire system to be created as non-transactional  for things to work)

In short: Express in MSMQ never made any difference. @ramonsmits do you remember where John Breakwells blow was recovered since he had a blogpost on this topic?

andreasohlund: Some info here https://docs.microsoft.com/en-gb/archive/blogs/johnbreakwell/why-are-my-express-msmq-messages-being-written-to-disk but I can't find the transactional queue parts

timbussmann:
As discussed with @andreasohlund , I've removed the express messages functionality completely now. Please have another look.

andreasohlund (approved):

PR #214, opened 2021-02-26
Update to NServiceBus Core alpha 852

timbussmann:
Adding the mandatory `CancellationToken` parameters and passing `CancellationToken.None` to all methods that require a cancellation token now. The actual usage of the tokens will be corrected by the related TF.

cc @andreasohlund


SzymonPobiega (approved):

timbussmann:
@andreasohlund just an FYI based on the last commit: there is a field named `cancellationToken` in the message pump class that is used to signal shutdown internally. Since the cancellation token parameter has the same name, it's easy to hide the field. That caused the shutdown to fail before the fix. Might be something to pay attention to when implementing the CTS usage.

PR #224, opened 2021-03-05
Update to NServiceBus Core alpha 880

timbussmann:
Implements the necessary changes to share the context with the `ErrorContext`. This needs to be carefully reviewed to ensure we use the correct context (current context for the incoming message or the stored context in the LRU storage).

I'm not sure if there are any potential side effects due to the (potentially very long) extended lifetime of the context on the `MessageContext` when storing state on the context.


andreasohlund:
https://github.com/Particular/NServiceBus.Transport.Msmq/pull/225 covers this and also makes the needed adjustments to support cancellation. I propose that we close this PR in favour of that one?

timbussmann:
👍

PR #225, opened 2021-03-06
Support cancellation

andreasohlund:
Updates to 8.0.0-alpha.880 and adjusts the transport to support cancalletion

### DIFF
--- a/src/NServiceBus.Transport.Msmq/MessagePump.cs
+++ b/src/NServiceBus.Transport.Msmq/MessagePump.cs
@@ -83,7 +83,7 @@ namespace NServiceBus.Transport.Msmq
             messageProcessingCancellationTokenSource = new CancellationTokenSource();
 
             // LongRunning is useless combined with async/await
-            messagePumpTask = Task.Run(() => ProcessMessages(messagePumpCancellationTokenSource.Token), cancellationToken);
+            messagePumpTask = Task.Run(() => ProcessMessages(), cancellationToken);
 
             return Task.CompletedTask;
         }
@@ -123,13 +123,13 @@ namespace NServiceBus.Transport.Msmq
         }
 
         [DebuggerNonUserCode]
-        async Task ProcessMessages(CancellationToken cancellationToken)
+        async Task ProcessMessages()
         {
-            while (!cancellationToken.IsCancellationRequested)
+            while (!messagePumpCancellationTokenSource.IsCancellationRequested)
             {
                 try
                 {
-                    await InnerProcessMessages(cancellationToken).ConfigureAwait(false);
+                    await InnerProcessMessages().ConfigureAwait(false);
                 }
                 catch (OperationCanceledException)
                 {
@@ -143,7 +143,7 @@ namespace NServiceBus.Transport.Msmq
             }
         }
 
-        async Task InnerProcessMessages(CancellationToken cancellationToken)
+        async Task InnerProcessMessages()
         {
             using (var enumerator = inputQueue.GetMessageEnumerator2())
             {
@@ -171,7 +171,7 @@ namespace NServiceBus.Transport.Msmq
                         return;
                     }
 
-                    await concurrencyLimiter.WaitAsync(cancellationToken).ConfigureAwait(false);
+                    await concurrencyLimiter.WaitAsync(messagePumpCancellationTokenSource.Token).ConfigureAwait(false);
 
                     _ = ReceiveMessage();
                 }
--- a/src/NServiceBus.Transport.Msmq/MessagePump.cs
+++ b/src/NServiceBus.Transport.Msmq/MessagePump.cs
@@ -188,6 +188,7 @@ namespace NServiceBus.Transport.Msmq
             return Task.Factory.StartNew(async state =>
             {
                 var accessStateForPerfReasons = (MessagePump)state;
+
                 try
                 {
                     await receiveStrategy.ReceiveMessage(messageProcessingCancellationTokenSource.Token).ConfigureAwait(false);
--- a/src/NServiceBus.Transport.Msmq/MessagePump.cs
+++ b/src/NServiceBus.Transport.Msmq/MessagePump.cs
@@ -83,7 +83,7 @@ namespace NServiceBus.Transport.Msmq
             messageProcessingCancellationTokenSource = new CancellationTokenSource();
 
             // LongRunning is useless combined with async/await
-            messagePumpTask = Task.Run(() => ProcessMessages(), cancellationToken);
+            messagePumpTask = Task.Run(() => PumpMessages(), cancellationToken);
 
             return Task.CompletedTask;
         }
@@ -123,13 +123,13 @@ namespace NServiceBus.Transport.Msmq
         }
 
         [DebuggerNonUserCode]
-        async Task ProcessMessages()
+        async Task PumpMessages()
         {
             while (!messagePumpCancellationTokenSource.IsCancellationRequested)
             {
                 try
                 {
-                    await InnerProcessMessages().ConfigureAwait(false);
+                    await InnerPumpMessages().ConfigureAwait(false);
                 }
                 catch (OperationCanceledException)
                 {
@@ -143,7 +143,7 @@ namespace NServiceBus.Transport.Msmq
             }
         }
 
-        async Task InnerProcessMessages()
+        async Task InnerPumpMessages()
         {
             using (var enumerator = inputQueue.GetMessageEnumerator2())
             {
--- a/src/NServiceBus.Transport.Msmq/MessagePump.cs
+++ b/src/NServiceBus.Transport.Msmq/MessagePump.cs
@@ -9,7 +9,7 @@ namespace NServiceBus.Transport.Msmq
     using Support;
     using Transport;
 
-    class MessagePump : IMessageReceiver, IDisposable
+    class MessagePump : IMessageReceiver
     {
         public MessagePump(
             Func<TransportTransactionMode, ReceiveStrategy> receiveStrategyFactory,
@@ -26,24 +26,15 @@ namespace NServiceBus.Transport.Msmq
             this.receiveSettings = receiveSettings;
         }
 
-        public void Dispose()
-        {
-            peekCircuitBreaker?.Dispose();
-            receiveCircuitBreaker?.Dispose();
-            messagePumpCancellationTokenSource?.Dispose();
-        }
-
         public Task Initialize(PushRuntimeSettings limitations, OnMessage onMessage, OnError onError, CancellationToken cancellationToken)
         {
             messagePumpCancellationTokenSource = new CancellationTokenSource();
             messageProcessingCancellationTokenSource = new CancellationTokenSource();
 
-            var tokenForCriticalErrorAction = messageProcessingCancellationTokenSource.Token; // Prevent ObjectDisposed after endpoint shut down by using a local variable
-
             peekCircuitBreaker = new RepeatedFailuresOverTimeCircuitBreaker("MsmqPeek", TimeSpan.FromSeconds(30),
-                ex => criticalErrorAction("Failed to peek " + receiveSettings.ReceiveAddress, ex, tokenForCriticalErrorAction));
+                ex => criticalErrorAction("Failed to peek " + receiveSettings.ReceiveAddress, ex, messageProcessingCancellationTokenSource.Token));
             receiveCircuitBreaker = new RepeatedFailuresOverTimeCircuitBreaker("MsmqReceive", TimeSpan.FromSeconds(30),
-                ex => criticalErrorAction("Failed to receive from " + receiveSettings.ReceiveAddress, ex, tokenForCriticalErrorAction));
+                ex => criticalErrorAction("Failed to receive from " + receiveSettings.ReceiveAddress, ex, messageProcessingCancellationTokenSource.Token));
 
             var inputAddress = MsmqAddress.Parse(receiveSettings.ReceiveAddress);
             var errorAddress = MsmqAddress.Parse(receiveSettings.ErrorQueue);
@@ -115,10 +106,11 @@ namespace NServiceBus.Transport.Msmq
                     .ConfigureAwait(false);
             }
 
-            concurrencyLimiter.Dispose();
-            inputQueue.Dispose();
-            errorQueue.Dispose();
-
+            concurrencyLimiter?.Dispose();
+            inputQueue?.Dispose();
+            errorQueue?.Dispose();
+            peekCircuitBreaker?.Dispose();
+            receiveCircuitBreaker?.Dispose();
             messagePumpCancellationTokenSource?.Dispose();
             messageProcessingCancellationTokenSource?.Dispose();
         }
--- a/src/NServiceBus.Transport.Msmq/MessagePump.cs
+++ b/src/NServiceBus.Transport.Msmq/MessagePump.cs
@@ -180,7 +180,7 @@ namespace NServiceBus.Transport.Msmq
 
         Task ReceiveMessage()
         {
-            return TaskEx.Run(async _ =>
+            return Task.Factory.StartNew(async _ =>
             {
                 try
                 {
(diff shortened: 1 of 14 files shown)


SzymonPobiega (approved):

src/NServiceBus.Transport.Msmq/MessagePump.cs
```diff
@@ -134,11 +143,11 @@ async Task ProcessMessages()
             }
         }
 
-        async Task InnerProcessMessages()
+        async Task InnerProcessMessages(CancellationToken cancellationToken)
```
timbussmann, line 146: would be easier to read if the `cancellationToken`'s naming would be more specific about which token this is. At first glance, it looks this is `messagePumpCancellationTokenSource` anyway and doesn't need to be passed (https://github.com/Particular/NServiceBus.Transport.Msmq/pull/225/files#diff-159e576e4eaa91dfa7405b89fc706862e8655e7c7c84ccc4aee0d5bdc8346098R86)?

src/NServiceBus.Transport.Msmq/NServiceBus.Transport.Msmq.csproj
```diff
@@ -14,7 +14,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="NServiceBus" Version="8.0.0-alpha.852" />
+    <PackageReference Include="NServiceBus" Version="8.0.0-alpha.880" />
```
timbussmann, line 17:
```suggestion
    <PackageReference Include="NServiceBus" Version="[8.0.0-alpha.880, 9.0.0)" />
```

src/NServiceBus.Transport.Msmq/MessagePump.cs
```diff
@@ -171,14 +180,12 @@ async Task InnerProcessMessages()
...
-                var messagePump = (MessagePump)state;
-
                 try
                 {
-                    await messagePump.receiveStrategy.ReceiveMessage().ConfigureAwait(false);
-                    messagePump.receiveCircuitBreaker.Success();
+                    await receiveStrategy.ReceiveMessage(messageProcessingCancellationTokenSource.Token).ConfigureAwait(false);
+                    receiveCircuitBreaker.Success();
                 }
                 catch (OperationCanceledException)
```
timbussmann, line 190: I noticed that David added an additional condition to these catches (https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/275/files#diff-bf785c1b37e39481e89856ea4195ca5c7e21f862b8eb2ffc43b63208a3c582e5R263). Would these make sense here as well?

andreasohlund: They are in the receive strategies, eg: https://github.com/Particular/NServiceBus.Transport.Msmq/pull/225/files#diff-4d6fd4708d86a05a95625c19f6da55d9d68be74498996e6a2aab8f90879a3a33R34 so I think that is already covered?

src/NServiceBus.Transport.Msmq/MessagePump.cs
andreasohlund: Agree, using messagePumpCancellationTokenSource  now

src/NServiceBus.Transport.Msmq/MessagePump.cs
```diff
@@ -171,14 +180,12 @@ async Task InnerProcessMessages()
 
         Task ReceiveMessage()
         {
-            return TaskEx.Run(async state =>
+            return Task.Run(async () =>
```
timbussmann, line 183: this seems to be a performance optimization to avoid variable capturing. Has this been intentionally removed?

andreasohlund: ah, just aligned with some of the other seams. I'll revert and add a comment

andreasohlund: Fixed

andreasohlund: or actually, does this work as intended with me inlining it?

src/NServiceBus.Transport.Msmq/MessagePump.cs
```diff
@@ -187,13 +194,19 @@ Task ReceiveMessage()
...
-                    messagePump.concurrencyLimiter.Release();
+                    concurrencyLimiter.Release();
                 }
-            }, this);
+            },
+            null,
+            CancellationToken.None,  // CancellationToken.None is used here since cancelling the task before it can run can cause the concurrencyLimiter to not be released
+            TaskCreationOptions.DenyChildAttach,
+            TaskScheduler.Default)
+                .Unwrap();
```
DavidBoike, line 208: I'm confused by this. MSMQ is the only transport I've seen do this. Should all transports be doing it?

andreasohlund: Not sure, @danielmarbach do U remember if we had some reasons for optimizing just MSMQ?

danielmarbach: Pretty sure we have done this to make sure there is less closure allocation. It was an important optimization at the time because a single receive operation is offloaded to the threadpool for every message that is handled. So if we manage to avoid capturing less in the closure (for example by allowing to remove the Func allocation or in the best case make sure there is no closure at all anymore) this adds up very quickly considering the hundreds of messages per second the transport can handle.

danielmarbach: Here is some more context. https://github.com/Particular/NServiceBus/issues/3884 As you can see Gen0 can be sometimes halved. But it really depends on the code the is underneath and you need to look at the decompiled code before and after

DavidBoike: Assuming this stays in MSMQ, it needs a Big Nasty Comment so we don't have to refer back to https://github.com/Particular/NServiceBus/issues/3884.

@danielmarbach Are you saying that if we take the compiled result of any given transport and decompile it, we'll be able to see whether or not the "Go process the message and we don't care about the returned task" operation results in a closure, and then if it does, we should be using this Task.Run construction, and if not, leave it alone?

danielmarbach: Very crude example since it is already late

```
    class Program
    {
        static async Task Main(string[] args)
        {
            var state = new object();

            await Captures(state);

            await DoesNotCapture(state);
        }

        static async Task Captures(object state) {
            await Task.Run(async () => {
                await Console.Error.WriteLineAsync(state.GetHashCode().ToString());
            });
        }

        static async Task DoesNotCapture(object state) {
            await Task.Factory.StartNew(async s => {
                await Console.Error.WriteLineAsync(s.GetHashCode().ToString());
            }, state,
            CancellationToken.None,
            TaskCreationOptions.DenyChildAttach,
            TaskScheduler.Default)
            .Unwrap();
        }
    }
```

First one leads to 

```
            Program.\u003C\u003Ec__DisplayClass1_0 cDisplayClass10 = new Program.\u003C\u003Ec__DisplayClass1_0();
            cDisplayClass10.state = this.state;
            // ISSUE: method pointer
            awaiter = Task.Run(new Func<Task>((object) cDisplayClass10, __methodptr(\u003CCaptures\u003Eb__0))).GetAwaiter();
```

vs

```
            awaiter = Task.Factory.StartNew<Task>(Program.\u003C\u003Ec.\u003C\u003E9__2_0 ?? (Program.\u003C\u003Ec.\u003C\u003E9__2_0 = new Func<object, Task>((object) Program.\u003C\u003Ec.\u003C\u003E9, __methodptr(\u003CDoesNotCapture\u003Eb__2_0))), this.state, CancellationToken.None, TaskCreationOptions.DenyChildAttach, TaskScheduler.Default).Unwrap().GetAwaiter();
```

To bring this into the per-message execution land the first one would create a display class and a Func allocation for every execution while the second one only statically allocates the Func ones and doesn't require a display class. So effectively two allocations per execution less. Though sometimes it is only possible to get rid of the Display class allocation and not the func allocation depending on the code you are trying to optimize.

src/NServiceBus.Transport.Msmq/MessagePump.cs
danielmarbach: The trick is to access the state and use that as a host to avoid the closure. So no most likely what you did there doesn't work as before

andreasohlund: Got it!  I've added the state access back and some comments. Does it look right?

src/NServiceBus.Transport.Msmq/MessagePump.cs
```diff
@@ -171,29 +182,38 @@ async Task InnerProcessMessages()
 
         Task ReceiveMessage()
         {
-            return TaskEx.Run(async state =>
+            // We pass a fake state to make sure we benefit from lamda delegate caching. See https://github.com/Particular/NServiceBus/issues/3884
+            var statePassedToAvoidLamdaDelegateCaching = this;
+
+            return Task.Factory.StartNew(async state =>
```
timbussmann, line 188: what's the reasoning to switch to all the Factory.StartNew that we didn't use before?

danielmarbach: It was used before see https://github.com/Particular/NServiceBus.Transport.Msmq/pull/225/files#diff-18f43e07cacce26b57ed472dfc89a38e8bfab0fb088fa76847b547b5eb4b25fdL37

timbussmann: ah, missed this 👍

src/NServiceBus.Transport.Msmq/MessagePump.cs
```diff
@@ -30,15 +30,19 @@ public void Dispose()
...
             receiveCircuitBreaker?.Dispose();
-            cancellationTokenSource?.Dispose();
+            messagePumpCancellationTokenSource?.Dispose();
         }
 
         public Task Initialize(PushRuntimeSettings limitations, OnMessage onMessage, OnError onError, CancellationToken cancellationToken)
         {
+            messagePumpCancellationTokenSource = new CancellationTokenSource();
+
+            var tokenForCriticalErrorAction = messageProcessingCancellationTokenSource.Token; // Prevent ObjectDisposed after endpoint shut down by using a local variable
```
timbussmann, line 40: `messageProcessingCancellationTokenSource` seems to be assigned here: https://github.com/Particular/NServiceBus.Transport.Msmq/pull/225/files#diff-159e576e4eaa91dfa7405b89fc706862e8655e7c7c84ccc4aee0d5bdc8346098R85 which is in `StartReceive`?

andreasohlund: Good catch, assignment moved the the correct place

src/NServiceBus.Transport.Msmq/MessagePump.cs
```diff
@@ -30,15 +30,20 @@ public void Dispose()
         {
             peekCircuitBreaker?.Dispose();
             receiveCircuitBreaker?.Dispose();
-            cancellationTokenSource?.Dispose();
+            messagePumpCancellationTokenSource?.Dispose();
```
timbussmann, line 33: is it intentional to dispose this twice? It seems to be already disposed in `StopReceive`.

src/NServiceBus.Transport.Msmq/MessagePump.cs
```diff
@@ -78,49 +83,54 @@ public Task StartReceive(CancellationToken cancellationToken)
...
+            messagePumpTask = Task.Run(() => PumpMessages(), cancellationToken);
 
             return Task.CompletedTask;
         }
 
         public async Task StopReceive(CancellationToken cancellationToken)
         {
-            cancellationTokenSource.Cancel();
+            messagePumpCancellationTokenSource?.Cancel();
+            cancellationToken.Register(() => messageProcessingCancellationTokenSource?.Cancel());
```
timbussmann, line 95: could this throw an `ObjectDisposedException` if the triggers after line 123? Seems like this could be avoided by setting the tokensource value to `null` after disposing it?

src/NServiceBus.Transport.Msmq/MessagePump.cs
```diff
@@ -171,29 +181,38 @@ async Task InnerProcessMessages()
 
         Task ReceiveMessage()
         {
-            return TaskEx.Run(async state =>
+            // We pass a fake state to make sure we benefit from lamda delegate caching. See https://github.com/Particular/NServiceBus/issues/3884
```
timbussmann, line 184:
```suggestion
            // We pass a state to make sure we benefit from lamda delegate caching. See https://github.com/Particular/NServiceBus/issues/3884
```

it's not really a "fake"?

src/NServiceBus.Transport.Msmq/MessagePump.cs
```diff
@@ -171,29 +181,38 @@ async Task InnerProcessMessages()
...
+            return Task.Factory.StartNew(async state =>
             {
-                var messagePump = (MessagePump)state;
+                var accessStateForPerfReasons = (MessagePump)state;
 
                 try
                 {
-                    await messagePump.receiveStrategy.ReceiveMessage().ConfigureAwait(false);
-                    messagePump.receiveCircuitBreaker.Success();
+                    await receiveStrategy.ReceiveMessage(messageProcessingCancellationTokenSource.Token).ConfigureAwait(false);
```
timbussmann, line 193: won't this cause variable capturing again, removing the benefit to passing the state?

src/NServiceBus.Transport.Msmq/MessagePump.cs
```diff
@@ -171,29 +181,38 @@ async Task InnerProcessMessages()
...
+                catch (OperationCanceledException) when (messageProcessingCancellationTokenSource.IsCancellationRequested)
                 {
-                    // Intentionally ignored
+                    // Graceful shutdown
                 }
                 catch (Exception ex)
                 {
                     Logger.Warn("MSMQ receive operation failed", ex);
-                    await messagePump.receiveCircuitBreaker.Failure(ex).ConfigureAwait(false);
+                    await receiveCircuitBreaker.Failure(ex).ConfigureAwait(false);
```
timbussmann, line 203: also variable capturing here

src/NServiceBus.Transport.Msmq/MessagePump.cs
```diff
@@ -171,29 +181,38 @@ async Task InnerProcessMessages()
...
                 }
                 finally
                 {
-                    messagePump.concurrencyLimiter.Release();
+                    concurrencyLimiter.Release();
```
timbussmann, line 207: and here?

src/NServiceBus.Transport.Msmq/MessagePump.cs
andreasohlund: totally missed that we implemented IDisposable, will fix

src/NServiceBus.Transport.Msmq/MessagePump.cs
```diff
@@ -157,43 +159,44 @@ async Task InnerProcessMessages()
...
-                    await concurrencyLimiter.WaitAsync(cancellationToken).ConfigureAwait(false);
+                    await concurrencyLimiter.WaitAsync(messagePumpCancellationTokenSource.Token).ConfigureAwait(false);
 
-                    _ = ReceiveMessage();
+                    _ = Task.Factory.StartNew(state => InnerReceiveMessages((MessagePump)state),
+                        this,  // We pass a state to make sure we benefit from lamda delegate caching. See https://github.com/Particular/NServiceBus/issues/3884
+                        CancellationToken.None,  // CancellationToken.None is used here since cancelling the task before it can run can cause the concurrencyLimiter to not be released
+                        TaskCreationOptions.DenyChildAttach,
+                        TaskScheduler.Default)
+                        .Unwrap();
```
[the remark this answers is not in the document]
danielmarbach: Strictly speaking, you wouldn't have to. My brain would stumble over it though because async body in `Task.Factory.StartNew` is always a dangerous territory. I would always prefer here the safe by default approach even though your point is absolutely valid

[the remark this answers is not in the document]
timbussmann: I'm fine either way. It seems the former (and now inlined) helper method correctly used `Unwrap` but in this specific case it wasn't used, so we could optimize it away. Anyway, no big deal.

DavidBoike (approved):

PR #242, opened 2021-03-19
Lazy initialization of subscriptions prevents potential race condition between retrieving subscription and publishing an event

SzymonPobiega:
Resolves #4164

### DIFF
--- a/src/NServiceBus.Transport.Msmq/Persistence/SubscriptionStorage/MsmqSubscriptionStorage.cs
+++ b/src/NServiceBus.Transport.Msmq/Persistence/SubscriptionStorage/MsmqSubscriptionStorage.cs
@@ -25,11 +25,6 @@ namespace NServiceBus.Persistence.Msmq
 
         void LoadSubscriptions()
         {
-            if (lookup != null)
-            {
-                return;
-            }
-
             lookup = new Dictionary<Subscriber, Dictionary<MessageType, string>>(SubscriberComparer);
 
             var messages = storageQueue.GetAllMessages()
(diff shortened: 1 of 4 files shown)


src/NServiceBus.Transport.Msmq/Persistence/SubscriptionStorage/MsmqSubscriptionStorage.cs
```diff
@@ -22,42 +22,40 @@ public void Dispose()
...
 
-        public void Init()
+        void EnsureInitialized()
         {
+            if (lookup != null)
+            {
+                return;
+            }
+
+            lookup = new Dictionary<Subscriber, Dictionary<MessageType, string>>(SubscriberComparer);
```
timbussmann, line 32: shouldn't this be assigned once fully initialized to prevent race conditions when the dictionary is assigned but not fully initialized and concurrent code would then pass the `if (lookup != null)` checks (e.g. here: https://github.com/Particular/NServiceBus.Transport.Msmq/pull/242/files#diff-4c2891fedff963c564638bcbe1df252790afba61e1606d42718a9b3ccdecc347R67)?

SzymonPobiega: The `EnsureInitialized` is always called in the context of an acquired writer lock so I think this is safe.

timbussmann: I think I was wondering whether callers would skip to acquire the writer lock because `lookup` wasn't null anymore (but `EnsureInitialized` didn't fully complete yet) and therefore try to access it without it being fully initialized?

SzymonPobiega: All usages of the lookup call `EnsureInitialized` first and this method is the only place that does the null-check, within a critical section.

ramonsmits (approved):

PR #246, opened 2021-03-22
Init subscriptions using Lazy<T>

ramonsmits:
Alternative to https://github.com/Particular/NServiceBus.Transport.Msmq/pull/242 , this uses Lazy<Y> and also documents why Lazy initialization is required.

### DIFF
--- a/src/NServiceBus.Transport.Msmq.Tests/Persistence/MsmqSubscriptionStorageTests.cs
+++ b/src/NServiceBus.Transport.Msmq.Tests/Persistence/MsmqSubscriptionStorageTests.cs
@@ -210,7 +210,6 @@
             };
 
             var queue1 = new FakeStorageQueue();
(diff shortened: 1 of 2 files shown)


SzymonPobiega:
Looks good although I don't like the fact that we're dealing with two locking structures here, the reader-writer lock _and_ the Lazy. I don't have a strong opinion though. @timbussmann maybe you could help us be a tie-breaker?

timbussmann:
>  I don't like the fact that we're dealing with two locking structures here, the reader-writer lock and the Lazy.

I agree with that, but I feel the approach with Lazy is still easier to read and more importantly, harder to break the code by missing some preliminary locking/initialization.

I don't have a strong opinion either, both approaches seem to be fairly similar with no major benefit of one over the other. I can toss a coin if you want ;)

SzymonPobiega:
@ramonsmits then if you feel stronger about your solution, let's go with yours :)

PR #272, opened 2021-04-28
[Spike] NServiceBus 8 - Scheduling Timeouts / Delayed Message with SQL as storage and using an out-of-process timeout manager

ramonsmits:
Approach using an external timeout manager similar as in the "old" days. Scheduling logic not part of the transport itself but does have the logic to route messages that have DelayDeliveryWith or DoNotDeliverBefore properties set to a "timeouts" queue.

This timeout queue corresponds to the out-of-process timeout manager. Multiple endpoints can send delayed messages to a single timeout manager or use use multiple. Most obvious configuration would likely be a single timeout manager per machine but could also use a remote queue to have a single timeout manager per system.


ramonsmits:
Rejected approach

PR #291, opened 2021-07-05
Rename timeout to delayed delivery plus some cleanup

SzymonPobiega:
- Rename remaining classes to use `DelayedDelivery` instead of `Timeout`
- Add configuration for the remaining circuit breakers
- Add missing XML comments
- Add/adjust namespaces to match the folder structure
- Move all public classes to `NServiceBus` namespace


src/NServiceBus.Transport.Msmq/DelayedDelivery/DelayedDeliverySettings.cs
```diff
@@ -0,0 +1,125 @@
...
+        /// <param name="timeToTriggerStoreCircuitBreaker">Time to wait before triggering the circuit breaker that monitors the storing of delayed messages in the database. Defaults to 30 seconds.</param>
+        /// <param name="timeToTriggerFetchCircuitBreaker">Time to wait before triggering the circuit breaker that monitors the fetching of due delayed messages from the database. Defaults to 30 seconds.</param>
+        /// <param name="timeToTriggerDispatchCircuitBreaker">Time to wait before triggering the circuit breaker that monitors the dispatching of due delayed messages to the destination. Defaults to 30 seconds.</param>
+        /// <param name="maximumRecoveryFailuresPerSecond">Maximum number of recovery failures per second that triggers the recovery circuit breaker. Defaults to 1/s.</param>
+        public DelayedDeliverySettings(IDelayedMessageStore delayedMessageStore,
+            int retries = 10,
+            TimeSpan? timeToTriggerStoreCircuitBreaker = null,
+            TimeSpan? timeToTriggerFetchCircuitBreaker = null,
+            TimeSpan? timeToTriggerDispatchCircuitBreaker = null,
+            int maximumRecoveryFailuresPerSecond = 1)
```
ramonsmits, line 102: I'm wondering... as these are also public properties and the args optional, should we just remove the args and have the user set them via the properties.

src/NServiceBus.Transport.Msmq/DelayedDelivery/DelayedMessage.cs
```diff
@@ -0,0 +1,41 @@
...
+    {
+        /// <summary>
+        /// Date and time the message is due.
+        /// </summary>
+        public DateTime Time { get; set; }
+
+        /// <summary>
+        /// Native message ID
+        /// </summary>
+        public string Id { get; set; }
```
ramonsmits, line 19: Now thinking about it, should this be `TransportMessageId` or `MessageId`?
```suggestion
        public string Id { get; set; }
```

Where would be want to apply idempotency? I think transport message ID is safest. Although that could result in having multiple "clones" with the same message ID but different transport ID's.

src/NServiceBus.Transport.Msmq/DelayedDelivery/DueDelayedMessagePoller.cs
```diff
@@ -0,0 +1,341 @@
...
+            txOption = transportTransactionMode == TransportTransactionMode.TransactionScope
+                ? TransactionScopeOption.Required
+                : TransactionScopeOption.RequiresNew;
+            this.delayedMessageStore = delayedMessageStore;
+            errorQueue = timeoutsErrorQueue;
+            this.faultMetadata = faultMetadata;
+            this.numberOfRetries = numberOfRetries;
+            this.dispatcher = dispatcher;
+            fetchCircuitBreaker = new RepeatedFailuresOverTimeCircuitBreaker("MsmqDelayedMessageFetch", timeToTriggerFetchCircuitBreaker,
+                ex => criticalErrorAction("Failed to fetch due delayed messages from the storage", ex, CancellationToken.None));
```
ramonsmits, line 37: Mmm, it feels CancellationToken.None is not the right value. Shouldn't this be forwarded based on hosting settings?

src/NServiceBus.Transport.Msmq/DelayedDelivery/DueDelayedMessagePoller.cs
```diff
@@ -0,0 +1,341 @@
...
+
+            dispatchCircuitBreaker = new RepeatedFailuresOverTimeCircuitBreaker("MsmqDelayedMessageDispatch", timeToTriggerDispatchCircuitBreaker,
+                ex => criticalErrorAction("Failed to dispatch delayed messages to destination", ex, CancellationToken.None));
+
+            failureHandlingCircuitBreaker = new FailureRateCircuitBreaker("MsmqDelayedMessageFailureHandling", maximumRecoveryFailuresPerSecond, ex => criticalErrorAction("Failed to execute error handling for delayed message forwarding", ex, CancellationToken.None));
```
ramonsmits, line 42: To copy the same stile as the other circuit breakers.
```suggestion
            failureHandlingCircuitBreaker = new FailureRateCircuitBreaker("MsmqDelayedMessageFailureHandling", maximumRecoveryFailuresPerSecond,
                ex => criticalErrorAction("Failed to execute error handling for delayed message forwarding", ex, CancellationToken.None));
```

src/NServiceBus.Transport.Msmq/DelayedDelivery/DueDelayedMessagePoller.cs
```diff
@@ -0,0 +1,341 @@
...
+            //If the next timeout is within a minute from now, trigger the poller
+            if (DateTimeOffset.UtcNow.Add(MaxSleepDuration) > timeoutTime)
+            {
+                //If there is something already in the queue we are fine.
+                signalQueue.Writer.TryWrite(true);
+            }
+        }
+
+        /// <summary>
+        /// This method does not accept a cancellation token because it need to finish all the tasks that have been started even if cancellation is under way
```
ramonsmits, line 73:
```suggestion
        /// This method does not accept a cancellation token because it needs to finish all the tasks that have been started even if cancellation is under way.
```

src/NServiceBus.Transport.Msmq/DelayedDelivery/DueDelayedMessagePoller.cs
```diff
@@ -0,0 +1,341 @@
...
+        {
+            try
+            {
+                while (!cancellationToken.IsCancellationRequested)
+                {
+#pragma warning disable PS0021 // Highlight when a try block passes multiple cancellation tokens
+                    try
+#pragma warning restore PS0021 // Highlight when a try block passes multiple cancellation tokens
+                    {
+                        var completionSource = new TaskCompletionSource<DateTimeOffset?>();
```
ramonsmits, line 105: This section could use a human readable explanation of what happens here.

src/NServiceBus.Transport.Msmq/DelayedDelivery/DueDelayedMessagePoller.cs
```diff
@@ -0,0 +1,341 @@
...
+                        var handleTask = Poll(completionSource, cancellationToken);
+                        await taskQueue.Writer.WriteAsync(handleTask, cancellationToken).ConfigureAwait(false);
+                        var nextPoll = await completionSource.Task.ConfigureAwait(false);
+
+                        if (nextPoll.HasValue)
+                        {
+                            using (var waitCancelled = new CancellationTokenSource())
+                            {
+                                var combinedSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, waitCancelled.Token, cancellationToken);
```
ramonsmits, line 114: Seems like a typo? As `cancellationToken` is passed twice?

src/NServiceBus.Transport.Msmq/DelayedDelivery/DueDelayedMessagePoller.cs
```diff
@@ -0,0 +1,341 @@
...
+                                var signalTask = WaitForSignal(combinedSource.Token);
+
+                                await Task.WhenAny(waitTask, signalTask).ConfigureAwait(false);
+                                waitCancelled.Cancel();
+                            }
+                        }
+                    }
+                    catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
+                    {
+                        //Shutting down
```
ramonsmits, line 126:
```suggestion
                        //Shutting down
                        Log.Debug("Stopping loop because of triggered shutdown.");
```

src/NServiceBus.Transport.Msmq/DelayedDelivery/DueDelayedMessagePoller.cs
```diff
@@ -0,0 +1,341 @@
...
+
+        async Task WaitForSignal(CancellationToken cancellationToken)
+        {
+            try
+            {
+                await signalQueue.Reader.ReadAsync(cancellationToken).ConfigureAwait(false);
+            }
+            catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
+            {
+                //Shutting down
```
ramonsmits, line 152:
```suggestion
                Log.Debug("Waiting for signal cancelled due to shutdown");
```

src/NServiceBus.Transport.Msmq/DelayedDelivery/DueDelayedMessagePoller.cs
```diff
@@ -0,0 +1,341 @@
...
+                {
+                    await TrySendDelayedMessageToErrorQueue(timeout, exception, cancellationToken).ConfigureAwait(false);
+                }
+            }
+            catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
+            {
+                //Shutting down
+                throw;
+            }
+            catch (Exception exception)
```
ramonsmits, line 215: If we are shutting down shouldn't this then be a `return`?

Otherwise, maybe an improved comment:
```suggestion
            // Prevent the `catch(Exception)` to do any tasks in case of shutdown.
            catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
            {
                throw;
            }
            catch (Exception exception)
```

ramonsmits, line 215: OR 

Not invoking `catch (Exception exception)` when cancelled.
```suggestion
            catch (Exception exception) when (!cancellationToken.IsCancellationRequested)
```

src/NServiceBus.Transport.Msmq/DelayedDelivery/DueDelayedMessagePoller.cs
```diff
@@ -0,0 +1,341 @@
...
+            {
+                Log.Error("Failure during timeout polling.", exception);
+                if (timeout != null)
+                {
+                    await dispatchCircuitBreaker.Failure(exception, cancellationToken).ConfigureAwait(false);
+                    using (var scope = new TransactionScope(TransactionScopeOption.RequiresNew, transactionOptions, TransactionScopeAsyncFlowOption.Enabled))
+                    {
+                        await delayedMessageStore.IncrementFailureCount(timeout, cancellationToken).ConfigureAwait(false);
+                        scope.Complete();
+
```
ramonsmits, line 225:
```suggestion
```

src/NServiceBus.Transport.Msmq/DelayedDelivery/DueDelayedMessagePoller.cs
```diff
@@ -0,0 +1,341 @@
...
+                Dictionary<string, string> headersAndProperties = MsmqUtilities.DeserializeMessageHeaders(timeout.Headers);
+
+                ExceptionHeaderHelper.SetExceptionHeaders(headersAndProperties, exception);
+
+                foreach (KeyValuePair<string, string> pair in faultMetadata)
+                {
+                    headersAndProperties[pair.Key] = pair.Value;
+                }
+
+                Log.Info($"Dispatch to error queue transaction = {txOption}, Enlist = {txOption == TransactionScopeOption.Required}");
```
ramonsmits, line 297:
```suggestion
                Log.InfoFormat("Move {0} to error queue", timeout.Id);
```

src/NServiceBus.Transport.Msmq/DelayedDelivery/DueDelayedMessagePoller.cs
```diff
@@ -0,0 +1,341 @@
...
+                    var transportOperation = new TransportOperation(outgoingMessage, new UnicastAddressTag(errorQueue));
+                    await dispatcher.Dispatch(new TransportOperations(transportOperation), transportTransaction, CancellationToken.None)
+                        .ConfigureAwait(false);
+
+                    transportTx.Complete();
+                }
+            }
+            catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
+            {
+                //Shutting down
```
ramonsmits, line 313:
```suggestion
                Log.Debug("Aborted sending delayed message to error queue due to cancellation.")
```

ramonsmits (changes requested):

src/NServiceBus.Transport.Msmq/DelayedDelivery/DelayedMessage.cs
lailabougria: MessageId seems to make more sense to me, especially to avoid duplicates?

src/NServiceBus.Transport.Msmq/DelayedDelivery/DueDelayedMessagePoller.cs
lailabougria: No, it's not, this is creating a combined source based on multiple tokens so that if one of them is canceled, the source is canceled too.

src/NServiceBus.Transport.Msmq/DelayedDelivery/DueDelayedMessagePoller.cs
lailabougria:
```suggestion
                        //Shutting down
                        Log.Debug("A shutdown was triggered, canceling the Loop operation.");
```

src/NServiceBus.Transport.Msmq/DelayedDelivery/DueDelayedMessagePoller.cs
lailabougria:
```suggestion
                //Shutting down
                 Log.Debug("A shutdown was triggered, canceling the WaitForSignal operation.");
```

src/NServiceBus.Transport.Msmq/DelayedDelivery/DueDelayedMessagePoller.cs
lailabougria: I'm on the fence here. If we return the OperationCanceledException is swallowed, could it be that it needs to bubble up?

src/NServiceBus.Transport.Msmq/DelayedDelivery/DelayedDeliverySettings.cs
SzymonPobiega: 👍 for that, it would make this class much simpler.

src/NServiceBus.Transport.Msmq/DelayedDelivery/DueDelayedMessagePoller.cs
SzymonPobiega: SQLT passes there a token that is used to stop its message pump so I guess we should pass the poller's `tokenStouce.Token` here.

src/NServiceBus.Transport.Msmq/DelayedDelivery/DueDelayedMessagePoller.cs
SzymonPobiega: 👍

src/NServiceBus.Transport.Msmq/DelayedDelivery/DueDelayedMessagePoller.cs
SzymonPobiega: Actually it is a mistake. I added it twice for some reason. I was probably really tired.

src/NServiceBus.Transport.Msmq/DelayedDelivery/DueDelayedMessagePoller.cs
SzymonPobiega: This actually had a misleading comment.

src/NServiceBus.Transport.Msmq/DelayedDelivery/DueDelayedMessagePoller.cs
SzymonPobiega: I based this code on https://docs.particular.net/nservicebus/cancellation-and-catching-exceptions I added some code to `AwaitHandleTasks` to handle this `throw` but you are right it could be replaced with a `return` and `AwaitHandleTasks` would not have to handle it. 

I am fine either way, I am not an expert in cancellation.

ramonsmits: @SzymonPobiega it just reads strange as that catch block only does a throw which would be the default when not having catch. Hence adding a comment to ensure the intent it easily understood.

ramonsmits: @lailabougria No, it would not be swallowed but just thrown as it isn't matching the generic Exception catch.

src/NServiceBus.Transport.Msmq/DelayedDelivery/DueDelayedMessagePoller.cs
```diff
@@ -102,16 +108,24 @@ async Task Loop(CancellationToken cancellationToken)
                     try
 #pragma warning restore PS0021 // Highlight when a try block passes multiple cancellation tokens
                     {
+
+                        //We create a TaskCompletionSource source here and pass it to spawned Poll task. Once the Poll task is done with fetching
```
ramonsmits, line 112: ❤️

src/NServiceBus.Transport.Msmq/DelayedDelivery/DueDelayedMessagePoller.cs
```diff
@@ -0,0 +1,356 @@
...
+            {
+                while (taskQueue.Reader.TryRead(out var task))
+                {
+                    try
+                    {
+                        await task.ConfigureAwait(false);
+                    }
+                    catch (OperationCanceledException)
+                    {
+                        //Shutting down
```
ramonsmits, line 90:
```suggestion
```

src/NServiceBus.Transport.Msmq/DelayedDelivery/DueDelayedMessagePoller.cs
```diff
@@ -0,0 +1,356 @@
...
+                                var signalTask = WaitForSignal(combinedSource.Token);
+
+                                await Task.WhenAny(waitTask, signalTask).ConfigureAwait(false);
+                                waitCancelled.Cancel();
+                            }
+                        }
+                    }
+                    catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
+                    {
+                        //Shutting down
```
ramonsmits, line 140:
```suggestion
```

src/NServiceBus.Transport.Msmq/DelayedDelivery/DueDelayedMessagePoller.cs
```diff
@@ -0,0 +1,356 @@
...
+
+        async Task WaitForSignal(CancellationToken cancellationToken)
+        {
+            try
+            {
+                await signalQueue.Reader.ReadAsync(cancellationToken).ConfigureAwait(false);
+            }
+            catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
+            {
+                //Shutting down or no signal arrived while the wait period was over
```
ramonsmits, line 167:
```suggestion
```

ramonsmits (approved):
Ok with the current state, could have a slightly neater code comment or catch refactoring but is just a tiny improvement.

PR #292, opened 2021-07-07
Remove caused at-most-once dispatch if implementation didn't support `TransactionScope`

ramonsmits:


src/NServiceBus.Transport.Msmq/DelayedDelivery/DueDelayedMessagePoller.cs
```diff
@@ -191,7 +191,13 @@ async Task Poll(TaskCompletionSource<DateTimeOffset?> result, CancellationToken
                     if (timeout != null)
                     {
                         result.SetResult(null);
-                        await HandleDueDelayedMessage(timeout, cancellationToken).ConfigureAwait(false);
+                        HandleDueDelayedMessage(timeout, cancellationToken);
+
+                        var success = await delayedMessageStore.Remove(timeout, cancellationToken).ConfigureAwait(false);
+                        if (!success)
+                        {
+                            Log.WarnFormat("Potential more-than-once dispatch as delayed message already removed from storage.");
```
SzymonPobiega, line 199: Do we want to warn here? That would mean that any storage that does not support transaction scope would be triggering warnings, even if user is using e.g. the outbox.

Maybe debug logging and warn only in `SendsAtomicWithReceive` and `TransactionScope` modes?

SzymonPobiega (approved):

src/NServiceBus.Transport.Msmq/DelayedDelivery/DueDelayedMessagePoller.cs
ramonsmits: After discussion we decided to keep the warn log level as this should not happen as we assume implementations will use locks or leases and this warn would indicate a non optimal implementation.

PR #330, opened 2021-11-03
Remove unused method

timbussmann:
stumbled upon this method during some review and noticed that it doesn't seem to be used at all anymore.


andreasohlund (approved):

ramonsmits:
@SzymonPobiega @lailabougria Seems we missed this...

PR #351, opened 2021-11-23
Add How to test locally section to readme

timbussmann:
To document that this now requires a SQL server to run the tests successfully.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,4 +11,5 @@ Documentation can be found at <https://docs.particular.net/transports/msmq/>.
 * MSMQ needs to be installed and enabled on the current machine.See the [MSMQ configuration documentation](https://docs.particular.net/transports/msmq/?version=msmqtransport_2#msmq-configuration) on how to install MSMQ.
 * A MSSQL Server needs to be accessible either via:
   * A local SQL Express installation containing a `nservicebus` database.
-  * A custom SQL Server connection string in provided via the `SqlServerConnectionString` environment variable
+  * A custom SQL Server connection string in provided via the `SqlServerConnectionString` environment variable.
+    * The SQL Server used must support DTC (e.g. container-hosted SQL Server instances are not working)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -5,3 +5,10 @@ MSMQ transport for NServiceBus
 NOTE: As Microsoft is not making MSMQ available for .NET Core, building new systems using MSMQ is not recommended. See https://particular.net/blog/msmq-is-dead
 
 Documentation can be found at <https://docs.particular.net/transports/msmq/>.
+
+## How to test locally
+
+* MSMQ needs to be installed and enabled on the current machine.See the [MSMQ configuration documentation](https://docs.particular.net/transports/msmq/?version=msmqtransport_2#msmq-configuration) on how to install MSMQ.
+* A MSSQL Server needs to be accessible either via:
+  * A local SQL Express installation containing a `nservicebus` database.
```
timbussmann, line 13: do you also need to create the database manually upfront or will SQL Express create it on the fly?

README.md
```diff
@@ -5,3 +5,10 @@ MSMQ transport for NServiceBus
...
+  * A custom SQL Server connection string in provided via the `SqlServerConnectionString` environment variable
```
timbussmann, line 14: it looks like this doesn't even seem to work with a remote SQL instance running on a container given the tests seem to rely on MSDTC, is that correct?

README.md
lailabougria: This should be created upfront, the [tables will be created](https://github.com/Particular/NServiceBus.Transport.Msmq/blob/328ca2c1863ed20cac20b51e8e8d88644d294cf7/src/NServiceBus.Transport.Msmq/DelayedDelivery/Sql/TimeoutTableCreator.cs), but the database is assumed to be available.

README.md
lailabougria: That's a good point, I hadn't considered that since I have a SQL express running which i am using

SzymonPobiega (approved):

lailabougria (approved):

PR #360, opened 2021-12-13
Implement concurrency changing

SzymonPobiega:
Part of https://github.com/Particular/DeveloperExperience/issues/1516

### DIFF
--- a/src/NServiceBus.Transport.Msmq/MessagePump.cs
+++ b/src/NServiceBus.Transport.Msmq/MessagePump.cs
@@ -297,7 +297,7 @@ namespace NServiceBus.Transport.Msmq
         CancellationTokenSource messagePumpCancellationTokenSource;
         CancellationTokenSource messageProcessingCancellationTokenSource;
         int maxConcurrency;
-        SemaphoreSlim concurrencyLimiter;
+        volatile SemaphoreSlim concurrencyLimiter;
         MessageQueue errorQueue;
         MessageQueue inputQueue;
 
--- a/src/NServiceBus.Transport.Msmq/MessagePump.cs
+++ b/src/NServiceBus.Transport.Msmq/MessagePump.cs
@@ -85,7 +85,7 @@ namespace NServiceBus.Transport.Msmq
             return Task.CompletedTask;
         }
 
-        public async Task ChangeConcurrency(PushRuntimeSettings newLimitations, CancellationToken cancellationToken = new CancellationToken())
+        public async Task ChangeConcurrency(PushRuntimeSettings newLimitations, CancellationToken cancellationToken = default)
         {
             var oldLimiter = concurrencyLimiter;
             var oldMaxConcurrency = maxConcurrency;
--- a/src/NServiceBus.Transport.Msmq/MessagePump.cs
+++ b/src/NServiceBus.Transport.Msmq/MessagePump.cs
@@ -203,10 +203,10 @@ namespace NServiceBus.Transport.Msmq
 
                     _ = Task.Factory.StartNew(state =>
                         {
-                            var (messagePump, cancellationToken) = ((MessagePump, CancellationToken))state;
-                            return ReceiveMessagesSwallowExceptionsAndReleaseConcurrencyLimiter(messagePump, localLimiter, cancellationToken);
+                            var (messagePump, limiter, cancellationToken) = ((MessagePump, SemaphoreSlim, CancellationToken))state;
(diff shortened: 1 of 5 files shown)


src/NServiceBus.Transport.Msmq/MessagePump.cs
```diff
@@ -85,6 +85,28 @@ public Task StartReceive(CancellationToken cancellationToken = default)
...
+                //Drain and dispose of the old semaphore
+                while (oldLimiter.CurrentCount != oldMaxConcurrency)
+                {
+                    await Task.Delay(50, cancellationToken).ConfigureAwait(false);
+                }
+                oldLimiter.Dispose();
+            }
+            catch (Exception ex) when (ex.IsCausedBy(cancellationToken))
+            {
+                //Ignore, we are stopping anyway
```
danielmarbach, line 106: This comment is not entirely correct in this context

src/NServiceBus.Transport.Msmq/MessagePump.cs
```diff
@@ -85,6 +85,28 @@ public Task StartReceive(CancellationToken cancellationToken = default)
             return Task.CompletedTask;
         }
 
+        public async Task ChangeConcurrency(PushRuntimeSettings newLimitations, CancellationToken cancellationToken = new CancellationToken())
```
danielmarbach, line 88: Why ` = new CancellationToken()`? Shouldn't this be `default`?

src/NServiceBus.Transport.Msmq/MessagePump.cs
```diff
@@ -175,12 +197,14 @@ async Task PumpMessages(CancellationToken messagePumpCancellationToken)
...
-                    await concurrencyLimiter.WaitAsync(messagePumpCancellationToken).ConfigureAwait(false);
+                    var localLimiter = concurrencyLimiter;
+
+                    await localLimiter.WaitAsync(messagePumpCancellationToken).ConfigureAwait(false);
 
                     _ = Task.Factory.StartNew(state =>
                         {
                             var (messagePump, cancellationToken) = ((MessagePump, CancellationToken))state;
-                            return ReceiveMessagesSwallowExceptionsAndReleaseConcurrencyLimiter(messagePump, cancellationToken);
+                            return ReceiveMessagesSwallowExceptionsAndReleaseConcurrencyLimiter(messagePump, localLimiter, cancellationToken);
```
danielmarbach, line 207: This is now capturing a closure which was previously avoided. To avoid the closure, you need to pass in the limiter via the value tuple and extract it inside the `StartNew`

But see below comment. Maybe this is not needed

src/NServiceBus.Transport.Msmq/MessagePump.cs
```diff
@@ -192,7 +216,7 @@ async Task PumpMessages(CancellationToken messagePumpCancellationToken)
         }
 
         // This is static to prevent the method from accessing fields in the pump since that causes variable capturing and cause extra allocations
-        static async Task ReceiveMessagesSwallowExceptionsAndReleaseConcurrencyLimiter(MessagePump messagePump, CancellationToken messagePumpCancellationToken)
+        static async Task ReceiveMessagesSwallowExceptionsAndReleaseConcurrencyLimiter(MessagePump messagePump, SemaphoreSlim localConcurrencyLimiter, CancellationToken messagePumpCancellationToken)
```
danielmarbach, line 219: `localConcurrencyLimiter` maybe is irrelevant. How about `concurrencyLimiter`. Also why is this change needed? Following the logic you could also assign extract it from the pump at the beginning, assign it to a local and release that one at the end couldn't you?

SzymonPobiega: I think it is needed. We need to release the same semaphore as we acquired in the previous method to make sure we drain all semaphores correctly. So I'll fix the passing of the semaphore to use the tuple.

src/NServiceBus.Transport.Msmq/MessagePump.cs
SzymonPobiega: Thanks, I'll fix it.

src/NServiceBus.Transport.Msmq/MessagePump.cs
SzymonPobiega: 👍

src/NServiceBus.Transport.Msmq/MessagePump.cs
SzymonPobiega: 👍

WojcikMike (approved):

PR #381, opened 2022-02-17
Use native TransactionManager.MaxTimeout API

ramonsmits:
Code read actual data from `machine.config` while there is a API for this. Not using the API can result in startup failures is max timeout is altered in any way like via reflection. For example, when using a workaround like:

```c#
SetTransactionManagerField("_cachedMaxTimeout", true);
SetTransactionManagerField("_maximumTimeout", timeout);

void SetTransactionManagerField(string fieldName, object value) =>
    typeof(TransactionManager)
        .GetField(fieldName, BindingFlags.NonPublic | BindingFlags.Static)
        .SetValue(null, value);
```


danielmarbach (approved):

src/NServiceBus.Transport.Msmq/MsmqScopeOptions.cs
```diff
@@ -13,9 +12,7 @@ public MsmqScopeOptions(TimeSpan? requestedTimeout = null, IsolationLevel? reque
             var isolationLevel = IsolationLevel.ReadCommitted;
             if (requestedTimeout.HasValue)
             {
-                var maxTimeout = GetMaxTimeout();
-
-                if (requestedTimeout.Value > maxTimeout)
+                if (requestedTimeout.Value > TransactionManager.MaximumTimeout)
                 {
                     throw new ConfigurationErrorsException(
```
danielmarbach, line 17: Nit: Technically we could now replace this exception as well since we don't read anything from the config anymore. Probably a simple general exception or an ArgumentOutOfRangeException would be more descriptive here

ramonsmits: I would even say that previously it should also have been an argument exception as the value to compare is with is from the arguments.

PR #411, opened 2022-04-12
Default options from operations in outgoing behaviors might not be applied

timbussmann:
related to https://github.com/Particular/NServiceBus/issues/6223

### Symptoms

The `UseJournalQueue` or `UseDeadLetterQueue` settings on message operation options classes (e.g. `SendOptions`, `PublishOptions`, etc.) might be applied to additional message operations that are invoked using custom behaviors that are registered in the outgoing pipeline. The options of the "outer" operation (the one that triggers the outgoing pipeline in the first place) might be applied if the "inner" operation (the one that is called from within a behavior in the outgoing pipeline)  does not explicitly specify the same setting and therefore intends to use the default setting values. The default settings are "hidden" by the "outer" operation settings.

### Who's affected

You are affected if all of the following conditions apply:
* Using a custom behavior in an [outgoing pipeline stage](https://docs.particular.net/nservicebus/pipeline/steps-stages-connectors#stages-outgoing-pipeline-stages) to send messages via `context.Send` or `context.Publish`.

### Root cause

This is caused by [this bug](https://github.com/Particular/NServiceBus/issues/6223) in NServiceBus. NServiceBus 7.5.2, 7.6.2, and 7.7.2 store these settings in a correctly isolated way but downstream packages need to use the newly introduced APIs from those patches.


cquirosj:
@timbussmann, heads up, temporarily used 7.7.1-beta.1 to see if CI completes. Should_apply_default_reply_in_inner_send is failing.

timbussmann:
that seems to be a failure due to a brittle assertion that doesn't take the transport address translation into account properly

cquirosj (approved):

src/NServiceBus.Transport.Msmq/Utils/ContextExtensions.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace NServiceBus.Transport.Msmq
+{
+    using Extensibility;
+
+    static class ContextExtensions
+    {
+        public static ReadOnlyContextBag GetOperationProperties(this ContextBag contextBag)
```
timbussmann, line 7:
```suggestion
        // Implements the 'ExtendableOptionsExtensions.GetOperationProperties' extension from Core to prevent this patch to force an update to Core 7.7 (as this has also been fixed on 7.5 and 7.6).
        public static ReadOnlyContextBag GetOperationProperties(this ContextBag contextBag)
```

timbussmann: @cquirosj I think this might be a helpful comment to add. If you agree please commit before merging the PR.

PR #413, opened 2022-04-21
adding a missing bracket

cquirosj:


bording (approved):
Good catch, looks like we missed a typo!

DavidBoike (approved):

